PR #2, opened 2024-08-06
fix: update check to verify string, otherwise it fails to verify

RafilxTenfen:


gitferry (approved):

RafilxTenfen:
from talking with @KonradStaniec 
We should not have this check for version v25 since the vigilante does use some deprecated calls like `DumpPrivKey` which is not available on version 25 

Any opinions on what to do here? @SebastianElvis @gitferry

SebastianElvis:
Hmm I'm trying out local deployments of all migrated repositories, and come across this error

```
goroutine 1 [running]:
github.com/babylonlabs-io/vigilante/cmd/vigilante/cmd.GetReporterCmd.func1(0xc00140e600?, {0x4336581?, 0x4?, 0x43364e5?})
	/go/src/github.com/babylonlabs-io/vigilante/cmd/vigilante/cmd/reporter.go:53 +0x61c
github.com/spf13/cobra.(*Command).execute(0xc0013f2900, {0xc000ce5760, 0x2, 0x2})
	/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:987 +0xaa3
github.com/spf13/cobra.(*Command).ExecuteC(0xc0013f2600)
	/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1115 +0x3ff
github.com/spf13/cobra.(*Command).Execute(0x0?)
	/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1039 +0x13
main.main()
	/go/src/github.com/babylonlabs-io/vigilante/cmd/vigilante/main.go:17 +0x1d
panic: failed to open BTC client: zmq is only supported by bitcoind, but got bitcoind v24.0.0-v25.0.0
```

Seems related to this fix?

btcclient/client_block_subscriber.go
```diff
@@ -56,7 +57,7 @@ func NewWithBlockSubscriber(cfg *config.BTCConfig, retrySleepTime, maxRetrySleep
 		if err != nil {
 			return nil, fmt.Errorf("failed to get BTC backend: %v", err)
 		}
-		if backend != rpcclient.BitcoindPost25 {
+		if !strings.EqualFold(backend.String(), rpcclient.BitcoindPost25.String()) {
 			return nil, fmt.Errorf("zmq is only supported by bitcoind, but got %v", backend)
 		}
```
SebastianElvis, line 62:
```suggestion
		if _, ok := backend.(rpcclient.BitcoindVersion); !ok {
			return nil, fmt.Errorf("zmq is only supported by bitcoind, but got %v", backend)
		}
```

SebastianElvis (approved):
Nice catch! Caught this when debugging local deployment as well

btcclient/client_block_subscriber.go
SebastianElvis: This check is to ensure we use bitcoind not btcd, since only bitcoind supports zmq that is necessary for the vigilante. So as long as the type matches it's okay. The version of bitcoind is not necessarily 25 or after

SebastianElvis: This check is to ensure we use bitcoind not btcd, since only bitcoind supports zmq that is necessary for the vigilante. So as long as the type matches it's okay. The version of bitcoind is not necessarily 25 or after

SebastianElvis, line 62: This check is to ensure we use bitcoind not btcd, since only bitcoind supports zmq that is necessary for the vigilante. So as long as the type matches it's okay. The version of bitcoind is not necessarily 25 or after

SebastianElvis: So can we do sth like

```
		if !strings.Contains(backend.String(), "bitcoind") {
			return nil, fmt.Errorf("zmq is only supported by bitcoind, but got %v", backend)
		}
```

RafilxTenfen: modified, a bit different, but tested in deployments :+1:

RafilxTenfen:
waiting for a second look @SebastianElvis

lesterli:
@RafilxTenfen This [check](https://github.com/babylonlabs-io/vigilante/blob/rafilx/fix-check-backend/btcclient/client_block_subscriber.go#L60-L62) is not unnecessary b/c the [switch statement](https://github.com/babylonlabs-io/vigilante/blob/rafilx/fix-check-backend/btcclient/client_block_subscriber.go#L38) already ensures we are in the `types.Bitcoind` case, and The `rpcclient.New()` function is called with specific config to `bitcoind`. Can we remove the redundant check to simplify the code?

Same for [the check](https://github.com/babylonlabs-io/vigilante/blob/rafilx/fix-check-backend/btcclient/client_block_subscriber.go#L104-L106) from the `btcd` case, redundant check and inconsistency issue that the error message states "websocket is only supported by btcd", but the check is for a specific version (`rpcclient.BtcdPost2401`).

If there's a specific reason to keep these check (e.g., to catch configuration mismatches), consider moving it to a separate validation function that's called before the switch statement.

RafilxTenfen:
> https://github.com/babylonlabs-io/vigilante/pull/2#issuecomment-2285359865
sounds good, it was removed already
closing

PR #6, opened 2024-08-07
Bump Babylon to version with hot fix for retries

KonradStaniec:


gitferry (approved):
Tested it locally and it worked

PR #9, opened 2024-08-12
chore(e2e): removing btcd

Lazar955:
This PR removes btcd for running e2e tests and uses bitcoind. Reference [issue](https://github.com/babylonchain/vigilante/issues/226)


e2etest/unbondingwatcher_e2e_test.go
```diff
@@ -107,22 +82,33 @@ func TestUnbondingWatcher(t *testing.T) {
...
-	mBlock := tm.MineBlockWithTxs(t, tm.RetrieveTransactionFromMempool(t, []*chainhash.Hash{&unbondingTxHash}))
-	require.Equal(t, 2, len(mBlock.Transactions))
+	require.Eventually(t, func() bool {
+		return len(tm.RetrieveTransactionFromMempool(t, []*chainhash.Hash{&unbondingTxHash})) == 1
+	}, eventuallyWaitTimeOut, eventuallyPollTime)
+
+	minedBlock := tm.mineBlock(t)
+	require.Equal(t, 2, len(minedBlock.Transactions))
+
+	//mBlock := tm.MineBlockWithTxs(t, tm.RetrieveTransactionFromMempool(t, []*chainhash.Hash{&unbondingTxHash}))
```
KonradStaniec, line 110: leftover commented out code

e2etest/test_manager_btcstaking.go
```diff
@@ -140,12 +137,20 @@ func (tm *TestManager) CreateBTCDelegation(
 	require.NoError(t, err)
 
 	// send staking tx to Bitcoin node's mempool
-	_, err = tm.BTCWalletClient.SendRawTransaction(stakingMsgTx, true)
+	_, err = tm.BTCClient.SendRawTransaction(stakingMsgTx, true)
 	require.NoError(t, err)
 
 	// mine a block with this tx, and insert it to Bitcoin / Babylon
-	mBlock := tm.MineBlockWithTxs(t, tm.RetrieveTransactionFromMempool(t, []*chainhash.Hash{stakingMsgTxHash}))
+	//mBlock := tm.MineBlockWithTxs(t, tm.RetrieveTransactionFromMempool(t, []*chainhash.Hash{stakingMsgTxHash}))
```
KonradStaniec, line 144: commented out code

e2etest/test_manager.go
```diff
@@ -35,148 +31,50 @@ var (
...
-	// by 1, and then use hdSeed || numTestInstances as the seed for generating keys
-	numTestInstances = 0
-
-	existingWalletFile = "wallet.db"
-	exisitngWalletPass = "pass"
-	walletTimeout      = 86400
-
 	eventuallyWaitTimeOut = 40 * time.Second
 	eventuallyPollTime    = 1 * time.Second
+	regtestParams         = &chaincfg.RegressionNetParams
```
KonradStaniec, line 36: I think we should remove:
```
netParams        = &chaincfg.SimNetParams
```
which are above and use only `regtestParams` in tests

e2etest/test_manager.go
```diff
@@ -35,148 +31,50 @@ var (
...
 	defaultConfig := config.DefaultConfig()
 	// Config setting necessary to connect btcd daemon
-	defaultConfig.BTC.NetParams = "simnet"
-	defaultConfig.BTC.Endpoint = "127.0.0.1:18556"
+	defaultConfig.BTC.NetParams = "regtest"
+	defaultConfig.BTC.Endpoint = "127.0.0.1:18443"
 	// Config setting necessary to connect btcwallet daemon
-	defaultConfig.BTC.BtcBackend = "btcd"
+	defaultConfig.BTC.BtcBackend = types.Bitcoind
 	defaultConfig.BTC.WalletEndpoint = "127.0.0.1:18554"
```
KonradStaniec, line 46: One follow up task we can have is to remove any references to btcd backend, for example:
- `.WalletEndpoint` config option makes sense only in case of btcd/btcdwallet combo, in case of bitcoind there will be always one endpoint for bitcoind note/wallet
- `.BTC.BtcBackend ` options now makes little sense as we intend to support one backed ie bitcoind

e2etest/test_manager.go
```diff
@@ -35,148 +31,50 @@ var (
...
-
-	return coinbaseKey, coinbaseAddr, nil
+	return defaultConfig
 }
 
 type TestManager struct {
-	MinerNode        *rpctest.Harness
+	TestRpcClient    *rpcclient.Client
+	BitcoindHandler  *BitcoindTestHandler
 	BtcWalletHandler *WalletHandler
```
KonradStaniec, line 59: this is not used

e2etest/test_manager.go
```diff
@@ -191,79 +89,47 @@ func initBTCClientWithSubscriber(t *testing.T, cfg *config.Config, rpcClient *rp
 // NOTE: if handlers.OnFilteredBlockConnected, handlers.OnFilteredBlockDisconnected
```
KonradStaniec, line 89: comment need update as we no longer have any handlers

KonradStaniec (commented):
few small comments, but in general looks good 

Few additional things:
- in our makefile we can now delete `btcd` and `btcwallet` from `test-e2e` command: https://github.com/babylonlabs-io/vigilante/blob/dev/Makefile#L55
- in tools.go we can delete the btcd and btcwallet import - https://github.com/babylonlabs-io/vigilante/blob/dev/tools/tools.go#L8
- in our second go.mod we can also delete those - https://github.com/babylonlabs-io/vigilante/blob/dev/tools/go.mod#L9

btcclient/client_block_subscriber.go
```diff
@@ -51,15 +51,6 @@ func NewWithBlockSubscriber(cfg *config.BTCConfig, retrySleepTime, maxRetrySleep
...
 
-		// ensure we are using bitcoind as Bitcoin node, as zmq is only supported by bitcoind
-		backend, err := rpcClient.BackendVersion()
-		if err != nil {
-			return nil, fmt.Errorf("failed to get BTC backend: %v", err)
-		}
-		if backend != rpcclient.BitcoindPost25 {
-			return nil, fmt.Errorf("zmq is only supported by bitcoind, but got %v", backend)
-		}
-
```
RafilxTenfen, line 62: why remove this check?

Lazar955: Might be wrong, but I don't think this check is necessary, we already check if we are using bitcoind in switch statement, and also bitcoind versions pre 25 also support zmq

RafilxTenfen: got it, the removal of dump private key will not be done in this PR, right? 
https://github.com/babylonlabs-io/vigilante/pull/2#issuecomment-2271712255

Lazar955: No in this on, will be a separate PR

e2etest/test_manager.go
Lazar955: Makes sense to be a separate PR, as the config is used in other places also

e2etest/slasher_e2e_test.go
```diff
@@ -239,19 +169,19 @@ func TestSlasher_SlashingUnbonding(t *testing.T) {
...
 
 	bsTracker := bst.NewBTCSTakingTracker(
 		tm.BTCClient,
 		backend,
 		tm.BabylonClient,
 		&bstCfg,
 		&commonCfg,
-		logger,
-		metrics,
+		rootLogger,
```
RafilxTenfen, line 183: maybe we don't need logs for tests
```suggestion
		zap.NewNop(),
```

e2etest/slasher_e2e_test.go
```diff
@@ -283,8 +213,14 @@ func TestSlasher_SlashingUnbonding(t *testing.T) {
 		_, err := tm.BTCClient.GetRawTransaction(unbondingSlashingMsgTxHash)
 		return err == nil
 	}, eventuallyWaitTimeOut, eventuallyPollTime)
+
 	// mine a block that includes slashing tx
-	tm.MineBlockWithTxs(t, tm.RetrieveTransactionFromMempool(t, []*chainhash.Hash{unbondingSlashingMsgTxHash}))
+	require.Eventually(t, func() bool {
```
RafilxTenfen, line 218: Would be nice to wrap up these eventually check into a function inside the test manager, but is only a suggestion

RafilxTenfen (approved):
a couple of suggestions, but LGTM :rocket:

KonradStaniec (approved):
LGTM!

PR #15, opened 2024-08-15
chore(submitter): remove dumpPrivKey

Lazar955:
This pull request removes the usage of the dumpPrivKey RPC method from the submitter and end-to-end (E2E) tests. The dumpPrivKey method is specific to legacy wallets and is no longer supported in the latest versions of bitcoind. Eliminating its usage is a necessary step for upgrading to newer versions of bitcoind, as its deprecation would otherwise hinder the upgrade process.

Updates the version of bitcoind to v27.0

References [issue](https://github.com/babylonchain/vigilante/issues/227).

### DIFF
--- a/e2etest/bitcoind_node_setup.go
+++ b/e2etest/bitcoind_node_setup.go
@@ -67,6 +67,7 @@ func (h *BitcoindTestHandler) Start() {
 	}, startTimeout, 500*time.Millisecond, "bitcoind did not start")
 }
 
+// GetBlockCount retrieves the current number of blocks in the blockchain from the Bitcoind.
 func (h *BitcoindTestHandler) GetBlockCount() (int, error) {
 	buff, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"getblockcount"})
 	if err != nil {
@@ -78,6 +79,7 @@ func (h *BitcoindTestHandler) GetBlockCount() (int, error) {
 	return strconv.Atoi(parsedBuffStr)
 }
 
+// GenerateBlocks mines a specified number of blocks in the Bitcoind.
 func (h *BitcoindTestHandler) GenerateBlocks(count int) *GenerateBlockResponse {
 	buff, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"-generate", fmt.Sprintf("%d", count)})
 	require.NoError(h.t, err)
@@ -89,10 +91,9 @@ func (h *BitcoindTestHandler) GenerateBlocks(count int) *GenerateBlockResponse {
 	return &response
 }
 
+// CreateWallet creates a new wallet with the specified name and passphrase in the Bitcoind
 func (h *BitcoindTestHandler) CreateWallet(walletName string, passphrase string) *CreateWalletResponse {
-	// last false on the list will create legacy wallet. This is needed, as currently
-	// we are signing all taproot transactions by dumping the private key and signing it
-	// on app level. Descriptor wallets do not allow dumping private keys.
+	// last arg is true which indicates we are using descriptor wallets they do not allow dumping private keys.
 	buff, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"createwallet", walletName, "false", "false", passphrase, "false", "true"})
 	require.NoError(h.t, err)
 
@@ -109,7 +110,7 @@ func (h *BitcoindTestHandler) InvalidateBlock(blockHash string) {
 	require.NoError(h.t, err)
 }
 
-// ImportDescriptors - todo
+// ImportDescriptors imports a given Bitcoin address descriptor into the Bitcoind
 func (h *BitcoindTestHandler) ImportDescriptors(descriptor string) {
 	_, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"importdescriptors", descriptor})
 	require.NoError(h.t, err)
--- a/e2etest/bitcoind_node_setup.go
+++ b/e2etest/bitcoind_node_setup.go
@@ -93,7 +93,7 @@ func (h *BitcoindTestHandler) CreateWallet(walletName string, passphrase string)
 	// last false on the list will create legacy wallet. This is needed, as currently
 	// we are signing all taproot transactions by dumping the private key and signing it
 	// on app level. Descriptor wallets do not allow dumping private keys.
-	buff, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"createwallet", walletName, "false", "false", passphrase, "false", "false"})
+	buff, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"createwallet", walletName, "false", "false", passphrase, "false", "true"})
 	require.NoError(h.t, err)
 
 	var response CreateWalletResponse
@@ -106,5 +106,11 @@ func (h *BitcoindTestHandler) CreateWallet(walletName string, passphrase string)
 // InvalidateBlock invalidates blocks starting from specified block hash
 func (h *BitcoindTestHandler) InvalidateBlock(blockHash string) {
 	_, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"invalidateblock", blockHash})
+	require.NoError(h.t, err)
+}
+
+// ImportDescriptors - todo
+func (h *BitcoindTestHandler) ImportDescriptors(descriptor string) {
+	_, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"importdescriptors", descriptor})
 	require.NoError(h.t, err)
 }
--- a/e2etest/test_manager.go
+++ b/e2etest/test_manager.go
@@ -245,7 +245,7 @@ func importPrivateKey(btcHandler *BitcoindTestHandler) (*btcec.PrivateKey, error
 		return nil, err
 	}
 
-	descriptor := fmt.Sprintf("combo(%s)#%s", wif.String(), "timestamp")
+	descriptor := fmt.Sprintf("combo(%s)", wif.String())
 
 	// Create the JSON descriptor object.
 	descJSON, err := json.Marshal([]map[string]interface{}{
--- a/e2etest/test_manager.go
+++ b/e2etest/test_manager.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"encoding/hex"
+	"encoding/json"
+	"fmt"
 	"go.uber.org/zap"
 	"testing"
 	"time"
@@ -87,7 +89,6 @@ func StartManager(t *testing.T, numMatureOutputsInWallet uint32) *TestManager {
 	btcHandler.Start()
 	passphrase := "pass"
 	_ = btcHandler.CreateWallet("default", passphrase)
-	blocksResponse := btcHandler.GenerateBlocks(int(numMatureOutputsInWallet))
 
 	cfg := defaultVigilanteConfig()
 
@@ -101,6 +102,13 @@ func StartManager(t *testing.T, numMatureOutputsInWallet uint32) *TestManager {
 		HTTPPostMode:         true,
 	}, nil)
 	require.NoError(t, err)
+
+	err = testRpcClient.WalletPassphrase(passphrase, 600)
+	require.NoError(t, err)
+
+	walletPrivKey, err := importPrivateKey(btcHandler)
+	require.NoError(t, err)
+	blocksResponse := btcHandler.GenerateBlocks(int(numMatureOutputsInWallet))
 
 	btcClient := initBTCClientWithSubscriber(t, cfg)
 
@@ -136,11 +144,8 @@ func StartManager(t *testing.T, numMatureOutputsInWallet uint32) *TestManager {
 		return true
 	}, eventuallyWaitTimeOut, eventuallyPollTime)
 
-	err = testRpcClient.WalletPassphrase(passphrase, 600)
-	require.NoError(t, err)
-
-	walletPrivKey, err := testRpcClient.DumpPrivKey(minerAddressDecoded)
-	require.NoError(t, err)
+	//walletPrivKey, err := testRpcClient.DumpPrivKey(minerAddressDecoded)
+	//require.NoError(t, err)
 
 	return &TestManager{
 		TestRpcClient:   testRpcClient,
@@ -149,7 +154,8 @@ func StartManager(t *testing.T, numMatureOutputsInWallet uint32) *TestManager {
 		BitcoindHandler: btcHandler,
 		BTCClient:       btcClient,
 		Config:          cfg,
-		WalletPrivKey:   walletPrivKey.PrivKey,
+		//WalletPrivKey:   walletPrivKey.PrivKey,
+		WalletPrivKey: walletPrivKey,
 	}
 }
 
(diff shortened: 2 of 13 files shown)


e2etest/bitcoind_node_setup.go
```diff
@@ -93,7 +93,7 @@ func (h *BitcoindTestHandler) CreateWallet(walletName string, passphrase string)
 	// last false on the list will create legacy wallet. This is needed, as currently
 	// we are signing all taproot transactions by dumping the private key and signing it
 	// on app level. Descriptor wallets do not allow dumping private keys.
-	buff, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"createwallet", walletName, "false", "false", passphrase, "false", "false"})
+	buff, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"createwallet", walletName, "false", "false", passphrase, "false", "true"})
```
Lazar955, line 96: Legacy wallets were "false"

submitter/relayer/utils.go
```diff
@@ -4,97 +4,37 @@ import (
...
-			utxo.ScriptPK,
-			txscript.SigHashAll,
-			privKey,
-			true,
-		)
-		if err != nil {
-			return nil, err
+// IndexOfTxOut searches for the first TxOut with a PkScript of the given length.
+// It returns the index of the first match and true if found, otherwise 0 and false.
+func IndexOfTxOut(outs []*wire.TxOut, searchLen int) (uint, error) {
```
Lazar955, line 28: When using fundRawTx in out txs sometimes changes the order, so we need to find the one with change address

e2etest/bitcoind_node_setup.go
```diff
@@ -89,11 +91,10 @@ func (h *BitcoindTestHandler) GenerateBlocks(count int) *GenerateBlockResponse {
...
 }
 
+// CreateWallet creates a new wallet with the specified name and passphrase in the Bitcoind
 func (h *BitcoindTestHandler) CreateWallet(walletName string, passphrase string) *CreateWalletResponse {
-	// last false on the list will create legacy wallet. This is needed, as currently
-	// we are signing all taproot transactions by dumping the private key and signing it
-	// on app level. Descriptor wallets do not allow dumping private keys.
-	buff, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"createwallet", walletName, "false", "false", passphrase, "false", "false"})
+	// last arg is true which indicates we are using descriptor wallets they do not allow dumping private keys.
+	buff, _, err := h.m.ExecBitcoindCliCmd(h.t, []string{"createwallet", walletName, "false", "false", passphrase, "false", "true"})
```
Lazar955, line 97: Legacy wallets were "false", true for descriptor wallets

submitter/relayer/relayer.go
```diff
@@ -25,6 +24,10 @@ import (
 	"github.com/babylonlabs-io/vigilante/types"
 )
 
+const (
+	addrSize = 34
```
Lazar955, line 28: size of the change address in txOut

e2etest/test_manager.go
```diff
@@ -227,3 +233,35 @@ func (tm *TestManager) CatchUpBTCLightClient(t *testing.T) {
...
+
+	// Create the JSON descriptor object.
+	descJSON, err := json.Marshal([]map[string]interface{}{
+		{
+			"desc":      descriptor,
+			"active":    true,
+			"timestamp": "now",
+			"label":     "test key",
+		},
+	})
```
Lazar955, line 258: A bit of docs on the topic:
https://bips.dev/384/
https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md

Basically `combo` allows us to import a key and handle multiple types of btc scripts with a single descriptor command.

 "now" as a timestamp tells Bitcoind to start scanning from the current blockchain height

submitter/relayer/relayer.go
```diff
@@ -386,44 +386,54 @@ func (rl *Relayer) buildTxWithData(
...
 		return nil, err
 	}
 	tx.AddTxOut(wire.NewTxOut(0, dataScript))
 
-	// build txout for change
-	changeAddr, err := rl.GetChangeAddress()
+	feeRate := btcutil.Amount(rl.getFeeRate()).ToBTC()
+	rawTxResult, err := rl.BTCWallet.FundRawTransaction(tx, btcjson.FundRawTransactionOpts{
+		FeeRate: &feeRate,
+		//EstimateMode: &btcjson.EstimateModeConservative,
```
KonradStaniec, line 400: commented out code

submitter/relayer/relayer.go
```diff
@@ -386,44 +386,54 @@ func (rl *Relayer) buildTxWithData(
...
 	dataScript, err := builder.AddOp(txscript.OP_RETURN).AddData(data).Script()
 	if err != nil {
 		return nil, err
 	}
 	tx.AddTxOut(wire.NewTxOut(0, dataScript))
 
-	// build txout for change
-	changeAddr, err := rl.GetChangeAddress()
+	feeRate := btcutil.Amount(rl.getFeeRate()).ToBTC()
+	rawTxResult, err := rl.BTCWallet.FundRawTransaction(tx, btcjson.FundRawTransactionOpts{
```
KonradStaniec, line 398: While looking at this function docs - https://developer.bitcoin.org/reference/rpc/fundrawtransaction.html#argument-2-options
it seems we can specify the position of change output in arguments - `changePosition`.

Maybe it would be better to use this option instead finding change output on our own ?

e2etest/test_manager.go
KonradStaniec: I would add this line with brief description above the line where we create descriptor

submitter/relayer/relayer.go
```diff
@@ -278,6 +273,7 @@ func (rl *Relayer) convertCkptToTwoTxAndSubmit(ckpt *ckpttypes.RawCheckpointResp
 
 	// this is to wait for btcwallet to update utxo database so that
 	// the tx that tx1 consumes will not appear in the next unspent txs lit
+	// todo: Lazar chech if this needed?
```
KonradStaniec, line 276: unnecessary comment ?

submitter/relayer/relayer.go
```diff
@@ -386,44 +386,54 @@ func (rl *Relayer) buildTxWithData(
 	rl.logger.Debugf("Building a BTC tx using %v with data %x", utxo.TxID.String(), data)
 	tx := wire.NewMsgTx(wire.TxVersion)
 
-	outPoint := wire.NewOutPoint(utxo.TxID, utxo.Vout)
```
KonradStaniec, line 389: the `utxo *types.UTXO` argument to the function is probably not needed now as all intput to the transaction as selected by `rl.BTCWallet.FundRawTransaction` function.

Given that we probably need two versions of this function:
1. first version for the first transaction, in which we allow wallet to chose inputs automatically
2. second version for the second transaction, in which we are able to specifiy input manually to use the transaction chaining

e2etest/test_manager.go
```diff
@@ -136,11 +144,8 @@ func StartManager(t *testing.T, numMatureOutputsInWallet uint32) *TestManager {
 		return true
 	}, eventuallyWaitTimeOut, eventuallyPollTime)
 
-	err = testRpcClient.WalletPassphrase(passphrase, 600)
-	require.NoError(t, err)
-
-	walletPrivKey, err := testRpcClient.DumpPrivKey(minerAddressDecoded)
-	require.NoError(t, err)
+	//walletPrivKey, err := testRpcClient.DumpPrivKey(minerAddressDecoded)
```
KonradStaniec, line 147: commented out code

RafilxTenfen (commented):
PR looks good, but what about [dumpPrivKeyAndSignTx](https://github.com/babylonlabs-io/vigilante/blob/d73b523c409a1444bb0ea94cf3dfa299d1a6e27a/submitter/relayer/relayer.go#L440) at the `processCheckpoints` case

RafilxTenfen (approved):
Code changes look good to me ^^

submitter/relayer/relayer.go
gitferry: Not sure how `FundRawTransaction` would simplify things as we cannot control the inputs used by `FundRawTransaction`. As such, for (1), how do we know the value of the output is sufficient to fund the second tx for chaining?

gitferry (commented):
Signing part lgtm! As per off-line discussion, my major concern is whether `FundRawTransaction` can handle the chaining without introducing much complexity. Maybe this part can be improved in a separate pr to not block here

submitter/relayer/relayer.go
KonradStaniec: If I understand our chaining mechanism correctly, want we want to do is:
1. send first transaction with pretty low fee rate, lets call it `Tx1`
2. then send the second transaction (`Tx2`)with higher fee rate and with input that is `Tx1` change output (this is necessary to achieve the chaining)

Imo `FundRawTransaction` can help here a lot and simplify stuff as:
1. For the `Tx1` we can just create unsigned and unfunded transaction with one output, and send it to `FundRawTransaction`, it will automatically choses inputs and add one change output which will have correct value to obey specified fee rate.
2. Then we can create `Tx2` with one input (change output from `Tx1` ) and one output and send it to `FundRawTransaction`. Now based on specified fee rate, endpoint will attach or not other inputs and will attach one change output to `Tx2` . This way chaining will be preserved,  and second tx will obey specified fee rate

gitferry: Oh, I didn't realize in (2) we can inject one input from the change output of tx1. If the case, then it indeed improves a lot. Sounds good to me!

KonradStaniec (approved):
We minor nits. Before merging lets wait for @gitferry approval as he is most familiar with this logic.

submitter/relayer/relayer.go
```diff
@@ -307,18 +282,10 @@ func (rl *Relayer) convertCkptToTwoTxAndSubmit(ckpt *ckpttypes.RawCheckpointResp
 
 // ChainTwoTxAndSend consumes one utxo and build two chaining txs:
 // the second tx consumes the output of the first tx
```
gitferry, line 293:
```suggestion
// ChainTwoTxAndSend builds two chaining txs with the given data
// the second tx consumes the output of the first tx
```

submitter/relayer/relayer.go
```diff
@@ -154,42 +159,39 @@ func (rl *Relayer) shouldResendCheckpoint(ckptInfo *types.CheckpointInfo, bumped
...
 // the result is multiplied by ResubmitFeeMultiplier set in config
 func (rl *Relayer) calculateBumpedFee(ckptInfo *types.CheckpointInfo) btcutil.Amount {
-	feeRate := rl.getFeeRate()
-	newTx1Fee := feeRate.FeeForVSize(ckptInfo.Tx1.Size)
-	newTx2Fee := feeRate.FeeForVSize(ckptInfo.Tx2.Size)
-	// minus the old fee of the first transaction because we do not want to pay again for the first transaction
-	bumpedFee := newTx1Fee + newTx2Fee - ckptInfo.Tx1.Fee
-
-	return bumpedFee.MulF64(float64(rl.config.ResubmitFeeMultiplier))
+	return ckptInfo.Tx2.Fee.MulF64(rl.config.ResubmitFeeMultiplier)
```
gitferry, line 162: why we don't want to estimate the fee here?

submitter/relayer/relayer.go
```diff
@@ -154,42 +159,39 @@ func (rl *Relayer) shouldResendCheckpoint(ckptInfo *types.CheckpointInfo, bumped
...
 }
 
 // resendSecondTxOfCheckpointToBTC resends the second tx of the checkpoint with bumpedFee
 func (rl *Relayer) resendSecondTxOfCheckpointToBTC(tx2 *types.BtcTxInfo, bumpedFee btcutil.Amount) (*types.BtcTxInfo, error) {
 	// set output value of the second tx to be the balance minus the bumped fee
 	// if the bumped fee is higher than the balance, then set the bumped fee to
 	// be equal to the balance to ensure the output value is not negative
```
gitferry, line 169: This seems outdated

submitter/relayer/relayer.go
```diff
@@ -154,42 +159,39 @@ func (rl *Relayer) shouldResendCheckpoint(ckptInfo *types.CheckpointInfo, bumped
...
-	balance := tx2.Utxo.Amount
+	balance := btcutil.Amount(tx2.Tx.TxOut[changePosition].Value)
+
+	// todo: revise this as this means we will end up with output with value 0 that will be rejected by bitcoind as dust output.
```
gitferry, line 172: can we make it an issue for reminder?

submitter/relayer/relayer.go
```diff
@@ -356,106 +321,100 @@ func (rl *Relayer) ChainTwoTxAndSend(
...
-	// record metrics of UTXOs' sum
-	rl.metrics.AvailableBTCBalance.Set(sum)
-
-	return utxo, nil
-}
-
 // buildTxWithData builds a tx with data inserted as OP_RETURN
 // note that OP_RETURN is set as the first output of the tx (index 0)
 // and the rest of the balance is sent to a new change address
 // as the second output with index 1
```
gitferry, line 327: this doc string needs update

submitter/relayer/relayer.go
```diff
@@ -356,106 +321,100 @@ func (rl *Relayer) ChainTwoTxAndSend(
...
 
-	outPoint := wire.NewOutPoint(utxo.TxID, utxo.Vout)
-	txIn := wire.NewTxIn(outPoint, nil, nil)
-	// Enable replace-by-fee
-	// See https://river.com/learn/terms/r/replace-by-fee-rbf
-	txIn.Sequence = math.MaxUint32 - 2
-	tx.AddTxIn(txIn)
+	if firstTx != nil {
+		txID := firstTx.TxHash()
+		outPoint := wire.NewOutPoint(&txID, 1)
```
gitferry, line 333: can we make `1` a self-explained constant?

submitter/relayer/relayer.go
```diff
@@ -356,106 +321,100 @@ func (rl *Relayer) ChainTwoTxAndSend(
...
 
 	// serialization
 	var signedTxBytes bytes.Buffer
-	err = tx.Serialize(&signedTxBytes)
-	if err != nil {
+	if err := tx.Serialize(&signedTxBytes); err != nil {
 		return nil, err
 	}
 
+	change := changeAmount - txFee
```
gitferry, line 410: should we also check the change is higher than dust?

submitter/relayer/relayer.go
```diff
@@ -356,106 +321,100 @@ func (rl *Relayer) ChainTwoTxAndSend(
...
+
 	rl.logger.Debugf("Successfully composed a BTC tx with balance of input: %v, "+
```
gitferry, line 412: We probably need to remove the balance of input as we don't specify it anymore. And we need to make it a single line as otherwise the values are not well translated

gitferry (commented):
Sorry for the delay. Some quick comments:

submitter/relayer/relayer.go
Lazar955: We've changed how we estimate fee, by using the fee value from the response of `FundRawTransaction`. This produces a different value (a bit higher fee) than we used to, which means when we estimate the fee here it wouldn't be the same way as we did, and in local tests resulting in a lower calculated `bumpedFee` needing a higher `ResubmitFeeMultiplier`. So I see it either as estimating fee ourselves or using the response from `FundRawTransaction`. Open to any other inputs, I might be missing something

gitferry: This makes sense. Thanks for the explanation!

PR #20, opened 2024-08-23
chore(btcclient): use notifier interface

Lazar955:
- Removes usage of btc client with subscription to zmq
- Removes proprietary client implementation of zmq subscription as we now rely on lnd's

References [issue](https://github.com/babylonchain/vigilante/issues/199)


RafilxTenfen:
Can we remove https://github.com/babylonlabs-io/vigilante/blob/998cee9868b68d5b70478f768d4355ab1bb541cc/btcclient/client_wallet.go#L111

```go
func (c *Client) DumpPrivKey(address btcutil.Address) (*btcutil.WIF, error) {
	return c.Client.DumpPrivKey(address)
}

```

Lazar955:
> Can we remove
> 
> https://github.com/babylonlabs-io/vigilante/blob/998cee9868b68d5b70478f768d4355ab1bb541cc/btcclient/client_wallet.go#L111
> 
> ```go
> func (c *Client) DumpPrivKey(address btcutil.Address) (*btcutil.WIF, error) {
> 	return c.Client.DumpPrivKey(address)
> }
> ```

Yep, I've opened an issue to cleanup btc client and btc wallet interaces, they can be merged. So will be done in a separate PR

RafilxTenfen (approved):
LGTM, tested the vigilante with bitcoind on deployments/local and was able to have healthy BTC delegations again :tada:

reporter/block_handler.go
```diff
@@ -2,35 +2,59 @@ package reporter
...
+
 			var errorRequiringBootstrap error
-			if event.EventType == types.BlockConnected {
-				errorRequiringBootstrap = r.handleConnectedBlocks(event)
-			} else if event.EventType == types.BlockDisconnected {
-				errorRequiringBootstrap = r.handleDisconnectedBlocks(event)
+			if !reorg {
+				errorRequiringBootstrap = r.handleConnectedBlocks(epoch.Height, epoch.BlockHeader)
+			} else {
+				errorRequiringBootstrap = r.handleDisconnectedBlocks(epoch.BlockHeader)
```
gitferry, line 51: It seems that we do the same chaining check (from L35 to L45) in both `handleDisconnectedBlocks` and `handleDisconnectedBlocks`. I believe we can combine the logic in one `handleNewBlock()` method. Basically, if the new block cannot connect to the cache tip, we bootstrap. Otherwise, we add it to the cache, no needing of separate logic

gitferry (commented):
Nice work! Generally, different vigilante programs share a lot of code. So one point of the refactoring is to extract common parts into a libary (e.g., lnd notifier) and reduce duplicate code as much as possible. This PR is a good step heading this direction. We can continue this in future prs

reporter/block_handler.go
KonradStaniec: oh that this good idea 👍 

My initial thought was that we could, translate new block  events into connected / disconnected  block events i.e
If we have the chain:
`H1 -> H2 -> H3 `, now the the reorg happens and the chain becomes `H1 - H2' - H3'`, and now we receive `H3'`  then we can:
1. Find the common ancestor between forks i.e `H1` 
2. Calculate that `H2` and `H3` were removed, so send to disconnection events starting from the top
3. Now send connection events for `H2` and `H3` 

Bootstrapping after reorg is a bit brutal i.e it is costly operation, but given that re-orgs are pretty rare imo it is possible route to take and if we encounter some problems with it we can always go for more fancy approach.

monitor/btcscanner/block_handler.go
```diff
@@ -3,38 +3,61 @@ package btcscanner
...
+		return
+	}
+
+	blockNotifier, err := bs.btcNotifier.RegisterBlockEpochNtfn(nil)
+	if err != nil {
+		bs.logger.Errorf("Failed registering block epoch notifier")
+		return
+	}
+
+	defer blockNotifier.Cancel()
```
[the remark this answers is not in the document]
KonradStaniec: Yup I would start notifier in start function

cmd/vigilante/cmd/monitor.go
```diff
@@ -80,8 +75,28 @@ func GetMonitorCmd() *cobra.Command {
 			// register monitor metrics
 			monitorMetrics := metrics.NewMonitorMetrics()
 
+			// create the chain notifier
+			btcParams, err := netparams.GetBTCParams(cfg.BTC.NetParams)
+			if err != nil {
+				panic(fmt.Errorf("failed to get BTC net params: %w", err))
+			}
+			btcCfg := btcclient.CfgToBtcNodeBackendConfig(cfg.BTC, "")
```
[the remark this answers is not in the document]
Lazar955: Yep, there's a new issue to remove btcd from repo (previously merged PR regarding that was to use it in e2e)

cmd/vigilante/cmd/btcstaking_tracker.go
```diff
@@ -61,26 +60,17 @@ func GetBTCStakingTracker() *cobra.Command {
...
-			btcParams, err := netparams.GetBTCParams(cfg.BTC.NetParams)
+			btcNotifier, err := btcclient.NewNodeBackendWithParams(cfg.BTC, "")
 			if err != nil {
-				panic(fmt.Errorf("failed to get BTC parameter: %w", err))
-			}
-			btcCfg := btcclient.CfgToBtcNodeBackendConfig(cfg.BTC, "") // we will read certifcates from file
-			btcNotifier, err := btcclient.NewNodeBackend(btcCfg, btcParams, &btcclient.EmptyHintCache{})
-			if err != nil {
-				panic(fmt.Errorf("failed to create btc chain notifier: %w", err))
+				panic(err)
```
gitferry, line 73:
```suggestion
				panic(fmt.Errorf("failed to create btc chain notifier: %w", err))
```

monitor/btcscanner/block_handler.go
```diff
@@ -3,75 +3,77 @@ package btcscanner
...
 	if parentHash != cacheTip.BlockHash() {
 		return errors.New("cache is not up-to-date")
 	}
 
+	// get the block from hash
+	blockHash := header.BlockHash()
+	ib, _, err := bs.BtcClient.GetBlockByHash(&blockHash)
+	if err != nil {
+		// failing to request the block, which means a bug
+		panic(err)
```
gitferry, line 74:
```suggestion
		// failing to request the block
		panic(fmt.Errorf("failed to request block by hash: %s", blockHash.String))
```

monitor/btcscanner/btc_scanner.go
```diff
@@ -87,14 +91,23 @@ func (bs *BtcScanner) Start() {
...
+	blockNotifier, err := bs.btcNotifier.RegisterBlockEpochNtfn(nil)
+	if err != nil {
+		bs.logger.Errorf("Failed registering block epoch notifier")
+		return
+	}
+
 	// start handling new blocks
 	bs.wg.Add(1)
-	go bs.blockEventHandler()
+	go bs.blockEventHandler(blockNotifier)
```
gitferry, line 110: I was worried about possible racing issue between the `Bootstrap` and `blockEventHandler` threads. They both produce confirmed blocks to the same channel, so they need to be sequencial. I can think of two options:
1. keep them as two separate go routines and use the `Synced` controller to ensure `blockEventHandler` happens after `Bootstrap`
2. combine them into a single goroutine and when creating the notifier

I'm in favor of option (2) in which we probably can remove the `Synced` controller. Nevertheless, after bootstrapping, we should be able to get the best known block and put it in `btcNotifier.RegisterBlockEpochNtfn(bestKnownBlock)`, so that we don't duplicate the chain tip

reporter/reporter.go
```diff
@@ -108,10 +110,21 @@ func (r *Reporter) Start() {
...
 
-	r.bootstrapWithRetries(false)
+	r.bootstrapWithRetries()
+
+	if err := r.btcNotifier.Start(); err != nil {
+		r.logger.Errorf("Failed starting notifier")
+		return
+	}
+
+	blockNotifier, err := r.btcNotifier.RegisterBlockEpochNtfn(nil)
```
gitferry, line 120: Can we get the best known block after bootstrapping and set it as the parameter? Otherwise, the notifier will immediately get the chain tip which could be duplicate with the one processed during bootstrap

gitferry (commented):
Much cleaner! One last concern is whether the btc scanner can output a canonical chain in expected sequence after bootstraping and event handler switch. It would be good to have a test to cover the cases where `bootstraping` -> `block handler`, `block handler` -> `reorg happens` -> `bootstraping` -> `block handler`. Currently we only have test for the bootstraping case. The tests can be done in this or separate pr

cmd/vigilante/cmd/btcstaking_tracker.go
Lazar955: errors are already wrapped in the `NewNodeBackendWithParams`

monitor/btcscanner/block_handler.go
Lazar955: good spot, typo in suggestion

gitferry (approved):
Thanks for all the changes. Lgtm! It would be nice to have comprehensive tests about boostraping and normal block events but can be done in a separate pr

PR #28, opened 2024-08-28
chore(relayer): prevent tx1 resubmission

Lazar955:
If tx2 fails and we are retrying, we shouldn't resend tx1 again.

Referencing [issue](https://github.com/babylonchain/vigilante/issues/67)


submitter/relayer/relayer.go
```diff
@@ -295,14 +301,21 @@ func (rl *Relayer) convertCkptToTwoTxAndSubmit(ckpt *ckpttypes.RawCheckpointResp
 func (rl *Relayer) ChainTwoTxAndSend(data1 []byte, data2 []byte) (*types.BtcTxInfo, *types.BtcTxInfo, error) {
 	// recipient is a change address that all the
 	// remaining balance of the utxo is sent to
-	tx1, err := rl.buildTxWithData(data1, nil)
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to add data to tx1: %w", err)
-	}
+	tx1 := rl.lastSubmittedCheckpoint.Tx1
+	// prevent resending tx1 if it was successful
+	if rl.lastSubmittedCheckpoint.Tx1 == nil {
```
Lazar955, line 306: Relevant changes here, cache the TX1 if successful

submitter/relayer/relayer.go
```diff
@@ -80,9 +80,15 @@ func (rl *Relayer) SendCheckpointToBTC(ckpt *ckpttypes.RawCheckpointWithMetaResp
...
 	}
 
-	if rl.lastSubmittedCheckpoint == nil || rl.lastSubmittedCheckpoint.Epoch < ckptEpoch {
+	if rl.lastSubmittedCheckpoint == nil ||
+		rl.lastSubmittedCheckpoint.Tx1 == nil ||
+		rl.lastSubmittedCheckpoint.Epoch < ckptEpoch {
 		rl.logger.Infof("Submitting a raw checkpoint for epoch %v for the first time", ckptEpoch)
 
+		if rl.lastSubmittedCheckpoint == nil {
+			rl.lastSubmittedCheckpoint = &types.CheckpointInfo{}
```
Lazar955, line 89: Init the `lastSubmittedCheckpoint`

btcclient/client_wallet.go
```diff
@@ -35,7 +35,7 @@ func NewWallet(cfg *config.BTCConfig, parentLogger *zap.Logger) (*Client, error)
 		HTTPPostMode: true,
 		User:         cfg.Username,
 		Pass:         cfg.Password,
-		DisableTLS:   cfg.DisableClientTLS,
```
RafilxTenfen: why remove from config?

RafilxTenfen: wait this was confusing

RafilxTenfen: not sure why this appeared to me as code change at first

submitter/relayer/relayer.go
RafilxTenfen: Can we add this to the default structure at `New( ` https://github.com/babylonlabs-io/vigilante/blob/c8a5c3638e13a024676b143dd84b8de33041433f/submitter/relayer/relayer.go#L47

and then simplify the check by removing `rl.lastSubmittedCheckpoint == nil`

submitter/relayer/relayer.go
RafilxTenfen: is there any chance for `rl.lastSubmittedCheckpoint.Tx1` be loaded with the tx1 sent from the previous epoch?

btcclient/client_wallet.go
Lazar955: This was in previous PR, basically we are not using it in bitcoind.

gitferry (commented):
I was thinking an alternative solution.
if we fail tx2 and retry the submission of both tx1 and tx2, tx1 would actually fail because the tx1 is already in mempool. In our impl, we would return error in this case so that tx2 will not be sent. So, I was thinking maybe we can achieve the goal by identifying the `already in mempool` error of tx1 and ignore it. Wdyt? Probably we need e2e to test this case

submitter/relayer/relayer.go
```diff
@@ -80,9 +80,15 @@ func (rl *Relayer) SendCheckpointToBTC(ckpt *ckpttypes.RawCheckpointWithMetaResp
 		return nil
 	}
 
-	if rl.lastSubmittedCheckpoint == nil || rl.lastSubmittedCheckpoint.Epoch < ckptEpoch {
+	if rl.lastSubmittedCheckpoint == nil ||
+		rl.lastSubmittedCheckpoint.Tx1 == nil ||
+		rl.lastSubmittedCheckpoint.Epoch < ckptEpoch {
 		rl.logger.Infof("Submitting a raw checkpoint for epoch %v for the first time", ckptEpoch)
 
+		if rl.lastSubmittedCheckpoint == nil {
```
gitferry, line 88: Can we separate the two cases? if `shouldSendCompleCheckpoint()` --> `convertCkptToTwoTxAndSubmit`, else if `shouldSendTx2()` --> `retrySendTx2`. I think this would be logically cleaner and some low-level funtions can be reused.

gitferry (commented):
Code-wise lgtm! My major comment is that we kinda mixed the logic of handling the two cases. It would be good to separate the two cases.

Also, better to have both unit and e2e tests for this but can be done in a separate pr

submitter/relayer/relayer.go
```diff
@@ -238,49 +256,72 @@ func (rl *Relayer) signTx(tx *wire.MsgTx) (*wire.MsgTx, error) {
...
-		return nil, err
-	}
+	// Return the encoded data
+	return data1, data2, nil
+}
 
+func (rl *Relayer) logAndRecordCheckpointMetrics(tx1, tx2 *types.BtcTxInfo, epochNum uint64) {
 	// this is to wait for btcwallet to update utxo database so that
 	// the tx that tx1 consumes will not appear in the next unspent txs lit
+	// todo(Lazar): is the arbitrary timeout here necessary?
```
gitferry, line 289: Maybe not after switching to using `FundRawTransaction`. Let's check it in e2e

submitter/relayer/relayer.go
```diff
@@ -290,34 +331,66 @@ func (rl *Relayer) convertCkptToTwoTxAndSubmit(ckpt *ckpttypes.RawCheckpointResp
...
-	// remaining balance of the utxo is sent to
-	tx1, err := rl.buildTxWithData(data1, nil)
+func (rl *Relayer) retrySendTx2(ckpt *ckpttypes.RawCheckpointResponse) (*types.CheckpointInfo, error) {
+	_, data2, err := rl.encodeCheckpointData(ckpt)
+	if err != nil {
+		return nil, err
+	}
+
+	tx1 := rl.lastSubmittedCheckpoint.Tx1
+	if tx1 == nil {
```
gitferry, line 341: better to have a doc string for this function saying that the tx1 should not be nil

submitter/relayer/relayer.go
```diff
@@ -80,7 +81,14 @@ func (rl *Relayer) SendCheckpointToBTC(ckpt *ckpttypes.RawCheckpointWithMetaResp
 		return nil
 	}
 
-	if rl.lastSubmittedCheckpoint == nil || rl.lastSubmittedCheckpoint.Epoch < ckptEpoch {
+	sendCompleteCkpt := rl.lastSubmittedCheckpoint.Tx1 == nil ||
+		rl.lastSubmittedCheckpoint.Epoch < ckptEpoch
```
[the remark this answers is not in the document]
Lazar955: > we should check whether rl.lastSubmittedCheckpoint is nil first. Otherwise, this might 

-> lastSubmittedCheckpoint is now initialized in the constructor so, it shouldn't panic

submitter/relayer/relayer.go
Lazar955: But this function uses tx1 implicitly through state, I'll add the doc string above, but I would live the check in

submitter/relayer/relayer.go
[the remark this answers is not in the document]
gitferry: You are right. All good then

submitter/relayer/relayer.go
gitferry: yep yep, keeping the check is good here

gitferry (approved):
Great work!

KonradStaniec (approved):
Code looks good ! (though lets wait @gitferry approval)

One question I have, do you guys think we should have some persistence/check on BTC  chain whether checkpoint was not submitted already ? 

Two separate cases I have in mind:
- we send checkpoint to the mempool but we crashed. Now after restart we will send both transactions once again which will lead to money loss. (as will use different inputs for those transactions, those will be different transactions with the same data)
- if there are multiple vigilante reporters in the network, the checkpoint may be already BTC chain but not yet submitted to Babylon. In this case we will also probably lose money.

KonradStaniec:
> This is a long-term thinking. Basically the submitter can have a separate goroutine looking for submitted checkpoints from new BTC blocks and decide whether the submit it
> 
> So maybe we can start by adding a db because for phase-2 we are probably still the only org running vigilantes?

Hmm agreed that solving the case `1` is more important. 

Even if we do not solve the case `2` before mainnet launch, if we solve case `1` our vigiliante loses have are constrained by 1 submission per epoch. This is not ideal but imo we can leave with it. 

So this is will be extension of this task if I am correct: https://github.com/babylonchain/vigilante/issues/156. We need persistent storage to:
- save bootstrapping time after recovery
- do not waste fees if we already submitted transactions for certain epochs (cc: @Lazar955 )

gitferry:
Yep, https://github.com/babylonchain/vigilante/issues/156 is raised mostly for the monitor bootstrapping but I guess the submitter also needs it

Issue #8, opened 2024-08-07
Stop relaying on Babylon retry library


KonradStaniec:
Currently vigilante program relies on `"github.com/babylonlabs-io/babylon/types/retry"` for retries. 

This approach is sub-optimal, and should be changed to use some better known library like `"github.com/avast/retry-go/v4"`

Lazar955:
Fixed in #29

PR #29, opened 2024-09-02
chore(*): use proper retry

Lazar955:
Replaces usage of `retry.do` from babylon with `avast/retry-go`

[References issue](https://github.com/babylonlabs-io/vigilante/issues/8)

### DIFF
--- a/btcclient/query.go
+++ b/btcclient/query.go
@@ -75,6 +75,7 @@ func (c *Client) getBestBlockHashWithRetry() (*chainhash.Hash, error) {
 	},
 		retry.Delay(c.retrySleepTime),
 		retry.MaxDelay(c.maxRetrySleepTime),
+		retry.Attempts(c.maxRetryTimes),
 	); err != nil {
 		c.logger.Debug(
 			"failed to query the best block hash", zap.Error(err))
@@ -98,6 +99,7 @@ func (c *Client) getBlockHashWithRetry(height uint64) (*chainhash.Hash, error) {
 	},
 		retry.Delay(c.retrySleepTime),
 		retry.MaxDelay(c.maxRetrySleepTime),
+		retry.Attempts(c.maxRetryTimes),
 	); err != nil {
 		c.logger.Debug(
 			"failed to query the block hash", zap.Uint64("height", height), zap.Error(err))
@@ -121,6 +123,7 @@ func (c *Client) getBlockWithRetry(hash *chainhash.Hash) (*wire.MsgBlock, error)
 	},
 		retry.Delay(c.retrySleepTime),
 		retry.MaxDelay(c.maxRetrySleepTime),
+		retry.Attempts(c.maxRetryTimes),
 	); err != nil {
 		c.logger.Debug(
 			"failed to query the block", zap.String("hash", hash.String()), zap.Error(err))
@@ -144,6 +147,7 @@ func (c *Client) getBlockVerboseWithRetry(hash *chainhash.Hash) (*btcjson.GetBlo
 	},
 		retry.Delay(c.retrySleepTime),
 		retry.MaxDelay(c.maxRetrySleepTime),
+		retry.Attempts(c.maxRetryTimes),
 	); err != nil {
 		c.logger.Debug(
 			"failed to query the block verbose", zap.String("hash", hash.String()), zap.Error(err))
--- a/btcclient/query.go
+++ b/btcclient/query.go
@@ -3,7 +3,7 @@ package btcclient
 import (
 	"fmt"
 
-	"github.com/babylonlabs-io/babylon/types/retry"
+	"github.com/avast/retry-go/v4"
 	"github.com/btcsuite/btcd/btcjson"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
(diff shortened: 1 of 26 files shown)


monitor/query.go
```diff
@@ -145,15 +157,25 @@ func (m *Monitor) queryEndedEpochBTCHeightWithRetry(epoch uint64) (*monitortypes
...
 	var reportedCheckpointBtcHeightRes monitortypes.QueryReportedCheckpointBtcHeightResponse
 
-	if err := retry.Do(m.ComCfg.RetrySleepTime, m.ComCfg.MaxRetrySleepTime, func() error {
+	retryOptions := []retry.Option{
+		retry.Delay(m.ComCfg.RetrySleepTime),
+		retry.MaxDelay(m.ComCfg.MaxRetrySleepTime),
+	}
+
+	if m.ComCfg.TestMode {
+		retryOptions = append(retryOptions, retry.Attempts(1))
```
Lazar955, line 166: ensure that functions in  `ReportedCheckpointBTCHeight` in FuzzLivenessChecker execute only once

KonradStaniec: maybe we could add `attempts` to config and set it to 1  in tests  ? Having special testmode in config is imo too hacky 😅

monitor/query.go
```diff
@@ -166,15 +182,18 @@ func (m *Monitor) queryReportedCheckpointBTCHeightWithRetry(hashStr string) (*mo
...
 		if err != nil {
 			return err
 		}
 
 		btcHeaderChainTipRes = *res
 		return nil
-	}); err != nil {
+	},
+		retry.Delay(m.ComCfg.RetrySleepTime),
+		retry.MaxDelay(m.ComCfg.MaxRetrySleepTime),
```
KonradStaniec, line 195: I would say we need to pass attempts config to every method, it would pretty wierd that some methods do not respect config value.

KonradStaniec (commented):
One note, babylon retry library have hardcoded errors which should/should not be retries:

```
var unrecoverableErrors = []error{
	btclctypes.ErrHeaderParentDoesNotExist,
	btclctypes.ErrChainWithNotEnoughWork,
	btclctypes.ErrInvalidHeader,
	btcctypes.ErrProvidedHeaderDoesNotHaveAncestor,
	btcctypes.ErrInvalidHeader,
	btcctypes.ErrNoCheckpointsForPreviousEpoch,
	btcctypes.ErrInvalidCheckpointProof,
	checkpointingtypes.ErrBlsPrivKeyDoesNotExist,
	// TODO Add more errors here
}

// expectedErrors is a list of errors which can safely be ignored and should not be retried.
var expectedErrors = []error{
	btcctypes.ErrDuplicatedSubmission,
	btcctypes.ErrInvalidHeader,
	// TODO Add more errors here
}
```

could you double check that this errors are relevant in this context ?

retrywrap/retry.go
```diff
@@ -0,0 +1,62 @@
...
+			return false
+		}
+
+		return true // Retry on all other errors
+	})
+
+	opts = append(opts, opt)
+
+	err := retry.Do(retryableFunc, opts...)
+	if containsErr(expectedErrors, err) {
```
RafilxTenfen, line 56: Why do we check the expected errors twice?

Lazar955: @RafilxTenfen when `retryif` returns false, that means that `Do` is returning an error, we want to check if this error is `expected` and ignore it (return nil), as we don't want to keep retrying on `expected` errors or `unrecoverable` but in that case we do return an error. This is matching behavior of retry implementation in Babylon repo

RafilxTenfen: got it, thanks for explaining

btcclient/query.go
```diff
@@ -66,13 +66,16 @@ func (c *Client) getBestBlockHashWithRetry() (*chainhash.Hash, error) {
...
-	if err := retry.Do(c.retrySleepTime, c.maxRetrySleepTime, func() error {
+	if err := retry.Do(func() error {
 		blockHash, err = c.GetBestBlockHash()
 		if err != nil {
 			return err
 		}
 		return nil
-	}); err != nil {
+	},
+		retry.Delay(c.retrySleepTime),
```
KonradStaniec, line 76: are those those options ever set in `btclient` ?

reporter/reporter.go
```diff
@@ -57,10 +58,13 @@ func New(
 		btccParamsRes *btcctypes.QueryParamsResponse
 		err           error
 	)
-	err = retry.Do(retrySleepTime, maxRetrySleepTime, func() error {
+	err = retrywrap.Do(func() error {
 		btccParamsRes, err = babylonClient.BTCCheckpointParams()
 		return err
-	})
+	},
+		retry.Delay(retrySleepTime),
```
KonradStaniec, line 65: I would say that, given `max-attempts` is in common potion of the config it should probably set in every program, otherwise we risk seeing some wierd behaviour only to learn some toold respect this config option and some does not

KonradStaniec (approved):
looks good, few minor comments.

One note for the future, for now it is fine to have this global list of errors, to not change behaviour, but ultimately this errors (unrecorevable/expected) should be set at call site by each caller. Then we know what call can return, what error and each caller knows what to do with each error. Maybe you could create task to get rid of those global error lists ?

reporter/reporter.go
Lazar955: Agreed added it in [f7c96e3](https://github.com/babylonlabs-io/vigilante/pull/29/commits/f7c96e39b7bc9ae631eee54b16d44d2e889fba60)

btcclient/query.go
Lazar955: Good catch, fixed in [f7c96e3](https://github.com/babylonlabs-io/vigilante/pull/29/commits/f7c96e39b7bc9ae631eee54b16d44d2e889fba60)

PR #34, opened 2024-09-09
feat(monitor): bootstrap monitor from db

Lazar955:
This PR enables the monitor to start from the last processed confirmed header, we also save the epoch

References [issue](https://github.com/babylonchain/vigilante/issues/156).


monitor/monitor.go
```diff
@@ -114,9 +124,30 @@ func (m *Monitor) Start() {
...
+	// get the height from db if it exists otherwise use the baseHeight provided from genesis
+	var startHeight uint64
+	dbHeight, exists, err := m.store.LatestHeight()
+	if err != nil {
+		panic(fmt.Errorf("error getting latest height from db %w", err))
+	} else if !exists {
+		startHeight = baseHeight
+	} else {
+		startHeight = dbHeight
+	}
```
RafilxTenfen, line 146: suggestion to extract this into a func

monitor/store/store.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (s *MonitorStore) put(key []byte, val uint64, bucketName []byte) error {
+	return kvdb.Batch(s.db, func(tx kvdb.RwTx) error {
+		bucket := tx.ReadWriteBucket(bucketName)
+		if bucket == nil {
+			return ErrCorruptedDb
+		}
+
+		return bucket.Put(key, uint64ToBytes(val))
+	})
```
RafilxTenfen, line 113:
```suggestion
	return kvdb.Update(s.db, func(tx kvdb.RwTx) error {
		bucket := tx.ReadWriteBucket(bucketName)
		if bucket == nil {
			return ErrCorruptedDb
		}

		return bucket.Put(key, uint64ToBytes(val))
	}, func() {})
```

monitor/store/store.go
```diff
@@ -0,0 +1,129 @@
...
+		returnVal = epoch
+
+		return nil
+	}, func() {
+	}); err != nil {
+		if errors.Is(err, ErrNotFound) {
+			return 0, false, nil
+		}
+		return 0, false, err
+	}
```
RafilxTenfen, line 100:
```suggestion
	err := s.db.View(func(tx walletdb.ReadTx) error {
		b := tx.ReadBucket(bucketName)
		if b == nil {
			return ErrCorruptedDb
		}

		value := b.Get(key)
		if value == nil {
			return ErrNotFound
		}

		epoch, err := uint64FromBytes(value)
		if err != nil {
			return err
		}
		returnVal = epoch

		return nil
	}, func() {})

	if err != nil {
		if errors.Is(err, ErrNotFound) {
			return 0, false, nil
		}
		return 0, false, err
	}
```

RafilxTenfen (approved):
Have you considered using the [`KVStore` ](https://github.com/cosmos/cosmos-sdk/blob/main/store/go.mod)to be similar to what is being used in the Babylon code?

besides that, only a few suggestions that are no impediments to merge! :tada:

Lazar955:
> Have you considered using the [`KVStore` ](https://github.com/cosmos/cosmos-sdk/blob/main/store/go.mod)to be similar to what is being used in the Babylon code?
> 
> besides that, only a few suggestions that are no impediments to merge! 🎉

Nice suggestion, but I've seen that we use this in our btc-staker and indexer, Konrad suggested to go with it

e2etest/test_manager.go
```diff
@@ -158,6 +159,8 @@ func (tm *TestManager) Stop(t *testing.T) {
 		err := tm.BabylonClient.Stop()
 		require.NoError(t, err)
 	}
+
+	//tm.BTCClient.Stop()
```
gitferry, line 163: leftover?

e2etest/test_manager.go
```diff
@@ -34,6 +34,7 @@ var (
 	eventuallyWaitTimeOut = 40 * time.Second
 	eventuallyPollTime    = 1 * time.Second
 	regtestParams         = &chaincfg.RegressionNetParams
+	defaultEpochInterval  = uint(400) //nolint:unused
```
gitferry, line 37:
```suggestion
	defaultEpochInterval  = uint(400)
```

e2etest/monitor_e2e_test.go
```diff
@@ -0,0 +1,157 @@
...
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		timer := time.NewTimer(15 * time.Second)
+		defer timer.Stop()
+
+		<-timer.C
+		mon.Stop()
+	}()
+	wg.Wait()
```
gitferry, line 129: Seems equivalent to:

```
time.Sleep(15 * time.Second)
mon.Stop()
```

gitferry (approved):
Great work! Some minor comments:

monitor/monitor.go
```diff
@@ -184,11 +221,15 @@ func (m *Monitor) handleNewConfirmedCheckpoint(ckpt *types.CheckpointRecord) err
...
 
 	nextEpochNum := m.GetCurrentEpoch() + 1
-	err = m.UpdateEpochInfo(nextEpochNum)
-	if err != nil {
+	if err := m.UpdateEpochInfo(nextEpochNum); err != nil {
 		return fmt.Errorf("failed to update information of epoch %d: %w", nextEpochNum, err)
 	}
 
+	// save the currently processed epoch
+	if err := m.store.PutLatestEpoch(m.GetCurrentEpoch()); err != nil {
```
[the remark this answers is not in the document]
Lazar955: Yeah, will save it in intermediary variable so it's clear, and start from +1 when loaded

Issue #23, opened 2024-08-23
Fix btcstacking-tracker race condition


Lazar955:
When running tests with `go test ./... -race -failfast` we get a race condition error:

```
WARNING: DATA RACE
Write at 0x00c005bb199c by goroutine 587:
  github.com/decred/dcrd/dcrec/secp256k1/v4.(*ModNScalar).NegateVal()
      /Users/lazar/go/pkg/mod/github.com/decred/dcrd/dcrec/secp256k1/v4@v4.2.0/modnscalar.go:1039 +0x388
  github.com/decred/dcrd/dcrec/secp256k1/v4.(*ModNScalar).Negate()
      /Users/lazar/go/pkg/mod/github.com/decred/dcrd/dcrec/secp256k1/v4@v4.2.0/modnscalar.go:1049 +0x1bc
  github.com/babylonlabs-io/babylon/crypto/schnorr-adaptor-signature.NewDecyptionKeyFromModNScalar()
      /Users/lazar/go/pkg/mod/github.com/babylonlabs-io/babylon@v0.9.1/crypto/schnorr-adaptor-signature/keys.go:26 +0x1b0
  github.com/babylonlabs-io/babylon/crypto/schnorr-adaptor-signature.NewDecyptionKeyFromBTCSK()
      /Users/lazar/go/pkg/mod/github.com/babylonlabs-io/babylon@v0.9.1/crypto/schnorr-adaptor-signature/keys.go:33 +0x60
  github.com/babylonlabs-io/babylon/x/btcstaking/types.(*BTCSlashingTx).BuildSlashingTxWithWitness()
      /Users/lazar/go/pkg/mod/github.com/babylonlabs-io/babylon@v0.9.1/x/btcstaking/types/btc_slashing_tx.go:280 +0x5c
  github.com/babylonlabs-io/vigilante/btcstaking-tracker/btcslasher.BuildSlashingTxWithWitness()
      /Users/lazar/work/vigilante/btcstaking-tracker/btcslasher/slasher_utils.go:353 +0x6c0
  github.com/babylonlabs-io/vigilante/btcstaking-tracker/btcslasher.(*BTCSlasher).sendSlashingTx()
      /Users/lazar/work/vigilante/btcstaking-tracker/btcslasher/slasher_utils.go:154 +0x294
  github.com/babylonlabs-io/vigilante/btcstaking-tracker/btcslasher.(*BTCSlasher).slashBTCDelegation.func1()
[...]

Issue #22, opened 2024-08-23
Make tests run with -race flag


Lazar955:
In order to prevent race condition all tests should run with `-race` flag

Lazar955:
done in #35

PR #35, opened 2024-09-10
fix(btcstaking-tracker): fix race condition

Lazar955:
- Fixes race condition where `*btcec.PrivateKey` is accessed from multiple go routines where read/write happen in `NewDecyptionKeyFromBTCSK` from pkg `github.com/babylonlabs-io/babylon/crypto/schnorr-adaptor-signature`
- Add `-race` flag to detect race conditions early
- Fixes [issue](https://github.com/babylonlabs-io/vigilante/issues/23) and references [issue](https://github.com/babylonlabs-io/vigilante/issues/22)


btcstaking-tracker/btcslasher/slasher.go
```diff
@@ -246,13 +246,18 @@ func (bs *BTCSlasher) SlashFinalityProvider(extractedfpBTCSK *btcec.PrivateKey)
 		return fmt.Errorf("failed to get BTC delegations under finality provider %s: %w", fpBTCPK.MarshalHex(), err)
 	}
 
+	// Initialize a mutex to protect access to extractedfpBTCSK *btcec.PrivateKey
+	var mu sync.Mutex
```
RafilxTenfen, line 250: if we need the mutex to protect access, maybe it is better to remove the wait group entirely 
or wrap the `extractedfpBTCSK *secp256k1.PrivateKey` into a protected structure

RafilxTenfen (approved):
:tada:

PR #43, opened 2024-09-13
feat(submitter): stateful submitter

Lazar955:
In case of a restart, we want to avoid wasting fees if we have already submitted transactions for a certain epoch.
Basic idea:
A crash occurs after sending both checkpoint transactions for epoch `n` to the BTC and recording them in the database. Upon restarting, we find that epoch `n` is still marked as sealed.

Before resending the transactions we first check our database and confirm that the transactions for this epoch have already been sent. Since the transactions were previously sent, the next step is to verify their status on our Bitcoin node.

If the Bitcoin node is aware of the transactions and they are present in the mempool, no further action is needed.
However, if the node does not recognize the transactions, this indicates they were lost, and we must resend them to the network.

[References](https://github.com/babylonlabs-io/vigilante/pull/28#issuecomment-2324805680)


submitter/relayer/relayer.go
```diff
@@ -542,3 +608,53 @@ func (rl *Relayer) sendTxToBTC(tx *wire.MsgTx) (*chainhash.Hash, error) {
 
 	return ha, nil
 }
+
+// maybeResendFromStore - checks if we need to resubmit txns from a store
+// in case "submitter" service was restarted, we want to ensure that we don't send txns again for a checkpoint
+// that has already been processed.
+// Returns true if the first transactions are in the mempool (no resubmission needed),
+// and false if any transaction was re-sent from the store.
+func maybeResendFromStore(
```
Lazar955, line 617: Mean and potato of the PR

submitter/relayer/relayer_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"errors"
+	"github.com/babylonlabs-io/vigilante/submitter/store"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func Test_maybeResendFromStore(t *testing.T) {
```
Lazar955, line 13: e2e test for this would be quite hard to implement with timings, opt for a unit

submitter/relayer/relayer.go
```diff
@@ -81,25 +95,68 @@ func (rl *Relayer) SendCheckpointToBTC(ckpt *ckpttypes.RawCheckpointWithMetaResp
...
+			rl.GetRawTransaction,
+			rl.sendTxToBTC,
+		)
+		if err != nil {
+			return err
+		}
+
+		if hasBeenProcessed {
+			return nil
+		}
```
gitferry, line 116: seems this chunk of code is irrelevant to whether we should send a complete ckpt or only the second half. Can we move it out to avoid duplicate code?

submitter/relayer/relayer.go
```diff
@@ -81,25 +95,68 @@ func (rl *Relayer) SendCheckpointToBTC(ckpt *ckpttypes.RawCheckpointWithMetaResp
...
 			return err
 		}
 
-		rl.lastSubmittedCheckpoint = submittedCheckpoint
+		rl.lastSubmittedCheckpoint = submittedCkpt
+
+		err = storeCkptFunc(submittedCkpt.Tx1.Tx, submittedCkpt.Tx2.Tx, submittedCkpt.Epoch)
+		if err != nil {
+			return err
+		}
```
gitferry, line 130: same here

gitferry (approved):
Great work! Minor comments:

PR #51, opened 2024-09-16
chore(btcclient): reduce rpc calls

Lazar955:
Reduces the number of RPC calls in methods by combining them.

[References issue](https://github.com/babylonchain/vigilante/issues/218)


RafilxTenfen (approved):
LGTM :tada:

Issue #25, opened 2024-08-26
E2E run babylon in docker


Lazar955:
- Currently in e2e tests, babylon is run in process with babylond command. A better and more consistent approach would be to use a docker container. This goes hand in hand with this [issue](https://github.com/babylonlabs-io/vigilante/issues/19)

Lazar955:
Done in #52

PR #52, opened 2024-09-17
chore(e2e): run babylond in docker

Lazar955:
Instead of managing babylond as a process, we run it in a Docker container. 
This will make e2e tests more consistent as bitcoind is also running in Docker and will enable us to run tests in parallel with `t.Parallel` (next PR).

[References issue](https://github.com/babylonlabs-io/vigilante/issues/25)


e2etest/container/container.go
```diff
@@ -160,6 +171,16 @@ func (m *Manager) RunBitcoindResource(
...
 			},
 		},
+		func(config *docker.HostConfig) {
+			config.PortBindings = map[docker.Port][]docker.PortBinding{
+				"8332/tcp":  {{HostIP: "", HostPort: strconv.Itoa(randomAvailablePort(t))}},
+				"8333/tcp":  {{HostIP: "", HostPort: strconv.Itoa(randomAvailablePort(t))}},
+				"28332/tcp": {{HostIP: "", HostPort: strconv.Itoa(randomAvailablePort(t))}},
+				"28333/tcp": {{HostIP: "", HostPort: strconv.Itoa(randomAvailablePort(t))}},
+				"18443/tcp": {{HostIP: "", HostPort: strconv.Itoa(randomAvailablePort(t))}},
+				"18444/tcp": {{HostIP: "", HostPort: strconv.Itoa(randomAvailablePort(t))}},
```
Lazar955, line 181: as we will be running multiple instances of both bitcoind and Babylon, ensure we use random ports so they don't clash

e2etest/test_manager.go
```diff
@@ -82,13 +84,19 @@ func initBTCClientWithSubscriber(t *testing.T, cfg *config.Config) *btcclient.Cl
...
+	btcHandler := NewBitcoindHandler(t, manager)
+	bitcoind := btcHandler.Start(t)
 	passphrase := "pass"
 	_ = btcHandler.CreateWallet("default", passphrase)
 
 	cfg := defaultVigilanteConfig()
 
+	cfg.BTC.Endpoint = fmt.Sprintf("127.0.0.1:%s", bitcoind.GetPort("18443/tcp"))
+	cfg.BTC.ZmqSeqEndpoint = fmt.Sprintf("tcp:// 127.0.0.1:%s", bitcoind.GetPort("28333/tcp"))
+
```
Lazar955, line 99: obtain relevant ports, based on the mapping of internal ones

e2etest/test_manager.go
```diff
@@ -121,16 +129,20 @@ func StartManager(t *testing.T, numMatureOutputsInWallet uint32, epochInterval u
...
-	require.NoError(t, err)
-	err = bh.Start()
+
+	tmpDir := t.TempDir()
+	babylond, err := manager.RunBabylondResource(t, tmpDir, baseHeaderHex, hex.EncodeToString(pkScript), epochInterval)
 	require.NoError(t, err)
 
 	// create Babylon client
-	cfg.Babylon.KeyDirectory = bh.GetNodeDataDir()
+	cfg.Babylon.KeyDirectory = filepath.Join(tmpDir, "node0", "babylond")
```
Lazar955, line 138: we expose this inside of our babylon container

e2etest/test_manager.go
```diff
@@ -121,16 +129,20 @@ func StartManager(t *testing.T, numMatureOutputsInWallet uint32, epochInterval u
...
 	cfg.Babylon.Key = "test-spending-key" // keyring to bbn node
 	cfg.Babylon.GasAdjustment = 3.0
 
+	// update port with the dynamically allocated one from docker
+	cfg.Babylon.RPCAddr = fmt.Sprintf("http://localhost:%s", babylond.GetPort("26657/tcp"))
+	cfg.Babylon.GRPCAddr = fmt.Sprintf("https://localhost:%s", babylond.GetPort("9090/tcp"))
+
```
Lazar955, line 145: get the assigned ports, same as above

RafilxTenfen (approved):
LGTM :tada:

e2etest/container/config.go
```diff
@@ -5,19 +5,24 @@ package container
...
+	BabylonRepository  string
+	BabylonVersion     string
 }
 
 //nolint:deadcode
 const (
 	dockerBitcoindRepository = "lncm/bitcoind"
 	dockerBitcoindVersionTag = "v27.0"
+	dockerBabylondRepository = "babylonlabs/babylond"
+	dockerBabylondVersionTag = "8e0222804ed19b18d74d599b80baa18f05e87d8a" // this is built from commit b1e255a
```
gitferry, line 17: so this should be manually updated to be consistent with the version in `go.mod`, correct? How to get the docker version?

e2etest/container/container.go
```diff
@@ -4,17 +4,31 @@ import (
...
 
 const (
-	bitcoindContainerName = "bitcoind-test"
+	bitcoindContainerName = "bitcoind"
+	babylondContainerName = "babylond"
+)
+
+var (
+	// jury
+	_, juryPK = btcec.PrivKeyFromBytes(
```
gitferry, line 29:
```suggestion
	_, covenantPK = btcec.PrivKeyFromBytes(
```

We don't use jury any more

e2etest/container/container.go
gitferry: just curious, why would we be running multiple instances? For e2e in vigilante, a single pair of bitcoind and babylond should be enough, no?

e2etest/container/container.go
```diff
@@ -169,20 +191,115 @@ func (m *Manager) RunBitcoindResource(
...
+
+	return resource, nil
+}
+
 // ClearResources removes all outstanding Docker resources created by the Manager.
 func (m *Manager) ClearResources() error {
 	for _, resource := range m.resources {
 		if err := m.pool.Purge(resource); err != nil {
-			return err
+			continue
```
gitferry, line 258: how do we know there's an error here?

e2etest/test_manager.go
```diff
@@ -82,13 +85,19 @@ func initBTCClientWithSubscriber(t *testing.T, cfg *config.Config) *btcclient.Cl
...
+	btcHandler := NewBitcoindHandler(t, manager)
+	bitcoind := btcHandler.Start(t)
 	passphrase := "pass"
 	_ = btcHandler.CreateWallet("default", passphrase)
 
 	cfg := defaultVigilanteConfig()
 
+	cfg.BTC.Endpoint = fmt.Sprintf("127.0.0.1:%s", bitcoind.GetPort("18443/tcp"))
+	cfg.BTC.ZmqSeqEndpoint = fmt.Sprintf("tcp:// 127.0.0.1:%s", bitcoind.GetPort("28333/tcp"))
```
gitferry, line 99: do we still support zmq?

gitferry (approved):
Nice work! Some questions:

gitferry:
Seems we can remove `tools/`?

e2etest/container/container.go
Lazar955: Possibly, but I feel like e2e tests should be isolated to confirm state, rather than tests possibly conflicting with one another

e2etest/container/config.go
Lazar955: It should be a tag, like v0.9.3, but our CI wasn't setup at the time to build for arm64. So yes, it should reflect go mod version

e2etest/test_manager.go
Lazar955: Yes, depending [on config](https://github.com/babylonlabs-io/vigilante/blob/49e8cb8297df6b3cb411ac1224c4c7588f85e264/btcclient/notifier.go#L118)

e2etest/container/container.go
Lazar955: This means that when we start running tests with `t.Parallel()` we will be spawning bitcoind and Babylon container per e2e tests, this ensures that ports won't conflict. There's no need for these tests to be sequential if we can run them in parallel

PR #54, opened 2024-09-18
chore(e2e): run e2e tests in parallel

Lazar955:
Execute e2e tests in parallel.
Speed up from `15m` -> `3m`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ test:
 	go test -race ./...
 
 test-e2e:
-	go test -mod=readonly --failfast -timeout=25m -v $(PACKAGES_E2E) -count=1 --parallel 6 --tags=e2e
+	go test -mod=readonly --failfast -timeout=25m -v $(PACKAGES_E2E) -count=1 --parallel 12 --tags=e2e
 
 build-docker:
 	$(DOCKER) build --tag babylonlabs-io/vigilante -f Dockerfile \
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -48,7 +48,7 @@ test:
 	go test -race ./...
 
 test-e2e:
-	go test -race -mod=readonly --failfast -timeout=25m -v $(PACKAGES_E2E) -count=1 --tags=e2e
+	go test -mod=readonly --failfast -timeout=25m -v $(PACKAGES_E2E) -count=1 --parallel 10 --tags=e2e
```
Lazar955, line 51: Removed the `-race` flag, as there were quite a bit of race conditions not related to the testing code, but some in Babylon some in cosmos SDK. Will open an issue to investigate this

-parallel 10 -> limits the number of concurrent tests to 10, we can play around with this, will depend on how much we can stress GH actions, or remove the limit

gitferry (approved):
This is great!

e2etest/container/container.go
```diff
@@ -222,20 +213,14 @@ func (m *Manager) RunBabylondResource(
 				fmt.Sprintf("%s/:/home/", mounthPath),
 			},
 			ExposedPorts: []string{
-				"1317",
-				"2345",
-				"9090",
-				"26656",
-				"26657",
+				"9090/tcp", // only expose what we need
```
RafilxTenfen, line 216: :100:

RafilxTenfen (approved):
:rocket: Awesome!

PR #57, opened 2024-09-18
chore(submitter): refactoring points 

Lazar955:
Code cleanup

[References issue ](https://github.com/babylonchain/vigilante/issues/178)

### DIFF
--- a/submitter/relayer/relayer.go
+++ b/submitter/relayer/relayer.go
@@ -10,7 +10,6 @@ import (
 	"github.com/lightningnetwork/lnd/kvdb"
 	"math"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/babylonlabs-io/babylon/btctxformatter"
@@ -296,20 +295,6 @@ func (rl *Relayer) resendSecondTxOfCheckpointToBTC(tx2 *types.BtcTxInfo, bumpedF
 	tx2.TxId = txID
 
 	return tx2, nil
-}
-
-func (rl *Relayer) inMempool() (bool, error) {
-	_, err := rl.GetRawTransaction(rl.lastSubmittedCheckpoint.Tx2.TxId)
-	if err != nil {
-		if strings.Contains(err.Error(), TxNotFoundErr.Error()) {
-			return false, nil
-		}
-		// Return the error if it's not a "not found" error
-		return false, err
-	}
-
-	// the transaction is in the mempool
-	return true, nil
 }
 
 // calcMinRelayFee returns the minimum transaction fee required for a
--- a/submitter/relayer/relayer.go
+++ b/submitter/relayer/relayer.go
@@ -259,9 +259,8 @@ func (rl *Relayer) resendSecondTxOfCheckpointToBTC(tx2 *types.BtcTxInfo, bumpedF
 	}
 
 	// No need to resend, transaction already confirmed
-	if status != btcclient.TxInMemPool && status != btcclient.TxNotFound {
-		rl.logger.Debugf("Transaction %v is already confirmed or has an unexpected state: %v",
-			rl.lastSubmittedCheckpoint.Tx2.TxId, status)
+	if status == btcclient.TxInChain {
+		rl.logger.Debugf("Transaction %v is already confirmed", rl.lastSubmittedCheckpoint.Tx2.TxId)
 		return nil, nil
 	}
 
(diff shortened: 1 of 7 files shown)


submitter/relayer/relayer.go
```diff
@@ -272,6 +288,44 @@ func (rl *Relayer) resendSecondTxOfCheckpointToBTC(tx2 *types.BtcTxInfo, bumpedF
 	return tx2, nil
 }
 
+func (rl *Relayer) checkFees() error {
```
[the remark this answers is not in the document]
Lazar955: Make sense, amended

submitter/relayer/relayer.go
```diff
@@ -240,6 +253,16 @@ func (rl *Relayer) calculateBumpedFee(ckptInfo *types.CheckpointInfo) btcutil.Am
 
 // resendSecondTxOfCheckpointToBTC resends the second tx of the checkpoint with bumpedFee
 func (rl *Relayer) resendSecondTxOfCheckpointToBTC(tx2 *types.BtcTxInfo, bumpedFee btcutil.Amount) (*types.BtcTxInfo, error) {
+	found, err := rl.inMempool()
+	if err != nil {
+		return nil, err
+	}
+
+	// No need to resend, transaction already confirmed
+	if !found {
```
RafilxTenfen, line 262: Is there any way that the Tx could be dropped from the mempool and it is not confirmed?

submitter/relayer/relayer.go
```diff
@@ -149,6 +154,18 @@ func (rl *Relayer) SendCheckpointToBTC(ckpt *ckpttypes.RawCheckpointWithMetaResp
 		return nil
 	}
 
+	return nil
+}
+
+// MaybeResubmitSecondCheckpointTx based on the resend interval attempts to resubmit 2nd ckpt tx with a bumped fee
+func (rl *Relayer) MaybeResubmitSecondCheckpointTx(ckpt *ckpttypes.RawCheckpointWithMetaResponse) error {
```
RafilxTenfen, line 161: only suggestion for naming
```suggestion
func (rl *Relayer) ResubmitSecondCheckpointTxIfNeeded(ckpt *ckpttypes.RawCheckpointWithMetaResponse) error {
```

submitter/relayer/relayer.go
KonradStaniec: if tx would have too low fee it is possible that it could be evicted from the mempool without inclusion in th ledger. Maybe it is worth adding the check if tx is is already in the ledger then ? In btc staker there is method - https://github.com/babylonchain/btc-staker/blob/main/walletcontroller/client.go#L243, which retrieves tx from node and marks wheter it is from the mempool or from the ledger. Maybe we can re-use it here ?

submitter/relayer/relayer.go
```diff
@@ -272,6 +295,20 @@ func (rl *Relayer) resendSecondTxOfCheckpointToBTC(tx2 *types.BtcTxInfo, bumpedF
...
+	_, err := rl.GetRawTransaction(rl.lastSubmittedCheckpoint.Tx2.TxId)
+	if err != nil {
+		if strings.Contains(err.Error(), TxNotFoundErr.Error()) {
+			return false, nil
+		}
+		// Return the error if it's not a "not found" error
+		return false, err
+	}
+
+	// the transaction is in the mempool
```
gitferry, line 308: here indicates that either the tx is either in the mempool or on the chain. I think we only need to resend tx2 if it is in the mempool or it can't be found at all. Can we differentiate it?

submitter/relayer/relayer.go
```diff
@@ -240,6 +252,19 @@ func (rl *Relayer) calculateBumpedFee(ckptInfo *types.CheckpointInfo) btcutil.Am
...
+	if err != nil {
+		return nil, err
+	}
+
+	// No need to resend, transaction already confirmed
+	if status != btcclient.TxInMemPool && status != btcclient.TxNotFound {
+		rl.logger.Debugf("Transaction %v is already confirmed or has an unexpected state: %v",
+			rl.lastSubmittedCheckpoint.Tx2.TxId, status)
+		return nil, nil
+	}
```
gitferry, line 266:
```suggestion
	if status == btcclient.TxInChain {
		rl.logger.Debugf("Transaction %v is already confirmed",
			rl.lastSubmittedCheckpoint.Tx2.TxId)
		return nil, nil
	}
```

Seems `TxInChain` is the only possible option here

gitferry (approved):
LGTM!

Issue #59, opened 2024-09-19
Improve randomPorts in e2e tests


Lazar955:
Sometimes the `randomAvailablePort` can assign the same port, improve this

PR #60, opened 2024-09-19
chore(e2e): better port selection

Lazar955:
Sometimes due to parallel tests, we could end up with the same port allocated. In this PR I try to introduce a small timeout when testing port availability by opening it for a short time which should decrease the chance of collision.

Tested with no collision couple of times with:
```sh
go test -mod=readonly -failfast -timeout=15m -v $(PACKAGES_E2E) -count=30 --parallel 12 --tags=e2e
```

[References issue](https://github.com/babylonlabs-io/vigilante/issues/59)

### DIFF
--- a/e2etest/container/container.go
+++ b/e2etest/container/container.go
@@ -5,8 +5,8 @@ import (
 	"context"
 	"fmt"
 	bbn "github.com/babylonlabs-io/babylon/types"
+	"github.com/babylonlabs-io/vigilante/testutil"
 	"github.com/btcsuite/btcd/btcec/v2"
-	"net"
 	"regexp"
 	"strconv"
 	"testing"
@@ -15,7 +15,6 @@ import (
 	"github.com/ory/dockertest/v3"
 	"github.com/ory/dockertest/v3/docker"
 	"github.com/stretchr/testify/require"
-	mrand "math/rand/v2"
 )
 
 const (
@@ -167,7 +166,7 @@ func (m *Manager) RunBitcoindResource(
 		},
 		func(config *docker.HostConfig) {
 			config.PortBindings = map[docker.Port][]docker.PortBinding{
-				"18443/tcp": {{HostIP: "", HostPort: strconv.Itoa(randomAvailablePort(t))}}, // only expose what we need
+				"18443/tcp": {{HostIP: "", HostPort: strconv.Itoa(testutil.AllocateUniquePort(t))}}, // only expose what we need
 			}
 			config.PublishAllPorts = false // because in dockerfile they already expose them
 		},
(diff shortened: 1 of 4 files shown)


e2etest/container/container.go
```diff
@@ -263,19 +262,20 @@ func randomAvailablePort(t *testing.T) int {
...
 	for i := 0; i < 10; i++ {
 		port := randPort(basePort, portRange)
 		listener, err := net.Listen("tcp", fmt.Sprintf("127.0.0.1:%d", port))
 		if err != nil {
 			continue
 		}
+
+		// listen on a port for a bit before closing, this should be just enough time for when other tests start
+		// when they try to net.Listen they should get an err and try a different port
+		time.Sleep(200 * time.Millisecond)
```
RafilxTenfen, line 278: this seems hacky... 
I would feel better by setting a specific port range for each test, but also doesn't seem too good 
Any other idea to solve this? @babylonlabs-io/core-dev

Lazar955: Right, but tests do not share state, so having a mutex or a map here doesn't help. One way would be to have a test which would be ran first (e.g called 01Setup and without parallel flag), which would create a file and a lockfile, then all other parallel tests startup and read/write port that they are acquiring to this shared file (locking and unlocking).
I feel like this is a bit much, but if we start having collisions I would implement it.
Thoughts @KonradStaniec @RafilxTenfen

Lazar955: Also reason for needing to have a test called 01Setup (or something like that), is that because all tests start in parallel so they would also be having race condition for the initial file creation

Lazar955: Not super hard to implement, I'll try it, see if it works and is worth the extra code. Thorugh this was clever and a nice hack :D

KonradStaniec: > Right, but tests do not share state, so having a mutex or a map here doesn't help.

They do not share it now but we could introduce it. You could create file `port_map.go` in our `e2etest` package have there `var port_map = map[int]bool` and protect access to it. It seems simpler that having file with lock file.

Or do I miss something about how go shares state between tests ?

testutil/port.go
```diff
@@ -0,0 +1,61 @@
...
+			continue
+		}
+
+		allocatedPorts[port] = struct{}{}
+		portMutex.Unlock()
+
+		if err := listener.Close(); err != nil {
+			continue
+		}
+
```
KonradStaniec, line 54: probably not critical right now becouse there is not much tests but we should also have here:

```
		t.Cleanup(func() {
			portMutex.Lock()
			defer portMutex.Unlock()
			delete(allocatedPorts, port)
		})
```

KonradStaniec (approved):

Issue #56, opened 2024-09-18
Flaky e2e tests investigate


Lazar955:
Investigate why TestReporter_BoostrapUnderFrequentBTCHeaders is very flaky

Lazar955:
fixed

PR #61, opened 2024-09-20
chore(e2e): fix flaky test

Lazar955:
Fixes flaky test when run in parallel

[References](https://github.com/babylonlabs-io/vigilante/issues/56)


Lazar955:
Seems like there is some conflict between TestHandleReorgAfterRestart and TestReporter_BoostrapUnderFrequentBTCHeaders

PR #68, opened 2024-09-30
fix: db config

RafilxTenfen:
- Fix db config mapping for `autocompactminage` and `dbtimeout`
- Add dbconfig to samples


Lazar955 (approved):
lgtm

PR #71, opened 2024-10-04
chore(e2e): e2e test for activating and unbonded in mempool

Lazar955:
This PR introduces an end-to-end test to validate that when both a staking transaction and an unbonding transaction are included in the same block, the delegation correctly transitions to the "UNBONDED" status.

### DIFF
--- a/btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
+++ b/btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
@@ -207,7 +207,7 @@ func (sew *StakingEventWatcher) fetchDelegations() {
 				continue
 			}
 
-			addToUnbonding := func(delegation Delegation) {
+			addToUnbondingFunc := func(delegation Delegation) {
 				del := &newDelegation{
 					stakingTxHash:         delegation.StakingTx.TxHash(),
 					stakingTx:             delegation.StakingTx,
@@ -216,14 +216,19 @@ func (sew *StakingEventWatcher) fetchDelegations() {
 					unbondingOutput:       delegation.UnbondingOutput,
 				}
 
-				// if we already have this delegation, skip it
+				// if we already have this delegation, we still want to check if it has changed,
 				// we should track both verified and active status for unbonding
-				if sew.unbondingTracker.GetDelegation(delegation.StakingTx.TxHash()) == nil {
+				changed, exists := sew.unbondingTracker.HasDelegationChanged(delegation.StakingTx.TxHash(), del)
+				if exists && changed {
+					// Delegation exists and has changed, push the update.
+					utils.PushOrQuit(sew.unbondingDelegationChan, del, sew.quit)
+				} else if !exists {
+					// Delegation doesn't exist, push the new delegation.
 					utils.PushOrQuit(sew.unbondingDelegationChan, del, sew.quit)
 				}
 			}
 
-			addToPending := func(delegation Delegation) {
+			addToPendingFunc := func(delegation Delegation) {
 				del := &newDelegation{
 					stakingTxHash:         delegation.StakingTx.TxHash(),
 					stakingTx:             delegation.StakingTx,
@@ -237,6 +242,8 @@ func (sew *StakingEventWatcher) fetchDelegations() {
 						del.stakingTx,
 						del.stakingOutputIdx,
 						del.unbondingOutput,
+						del.delegationStartHeight,
+						false,
 					)
 				}
 			}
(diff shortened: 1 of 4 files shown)


e2etest/unbondingwatcher_e2e_test.go
```diff
@@ -170,3 +199,117 @@ func TestActivatingDelegation(t *testing.T) {
 		return resp.BtcDelegation.Active
 	}, eventuallyWaitTimeOut, eventuallyPollTime)
 }
+
+// TestActivatingAndUnbondingDelegation tests that delegation will eventually become UNBONDED given that
+// both staking and unbonding tx are in the same block.
+// In this test, we include both staking tx and unbonding tx in the same block.
+// The delegation goes through "VERIFIED" → "ACTIVE" → "UNBONDED" status throughout this test.
+func TestActivatingAndUnbondingDelegation(t *testing.T) {
+	//t.Parallel()
```
Lazar955, line 208: for some reason, there might be an unwanted dependancy between e2e test, making it sequential until, will investigate (related to btc frequent headers maybe)

btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
```diff
@@ -216,14 +216,19 @@ func (sew *StakingEventWatcher) fetchDelegations() {
 					unbondingOutput:       delegation.UnbondingOutput,
 				}
 
-				// if we already have this delegation, skip it
+				// if we already have this delegation, we still want to check if it has changed,
 				// we should track both verified and active status for unbonding
-				if sew.unbondingTracker.GetDelegation(delegation.StakingTx.TxHash()) == nil {
+				changed, exists := sew.unbondingTracker.HasDelegationChanged(delegation.StakingTx.TxHash(), del)
```
Lazar955, line 221: the delegation might have changed from verified to active status, in order to track undonging, we need the activation height, and data in the tracker might be stale, so let's check this

e2etest/unbondingwatcher_e2e_test.go
Lazar955: "works on my machine" type of thing :sweat_smile:

btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
RafilxTenfen: this was a bit confusing the checks `exists && changed` or `!exists`

If the delegation didn't exists, we still push it as a new delegation to the `unbondingDelegationChan`?

btcstaking-tracker/stakingeventwatcher/tracked_delegations.go
```diff
@@ -86,3 +96,28 @@ func (td *TrackedDelegations) RemoveDelegation(stakingTxHash chainhash.Hash) {
 
 	delete(td.mapping, stakingTxHash)
 }
+
+func (td *TrackedDelegations) HasDelegationChanged(
+	stakingTxHash chainhash.Hash,
+	newDelegation *newDelegation,
+) (bool, bool) {
```
RafilxTenfen, line 103: suggestion to name returns 
```suggestion
) (exists bool, statusUpdated bool) {
```

btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
Lazar955: Yes, we want to track it, but we also want to push it to the channel if it exits and has changed (e.g went from verified -> activated)

RafilxTenfen (approved):
LGTM then ^^

btcstaking-tracker/stakingeventwatcher/tracked_delegations.go
```diff
@@ -86,3 +96,28 @@ func (td *TrackedDelegations) RemoveDelegation(stakingTxHash chainhash.Hash) {
...
+	// Check if the delegation exists in the map
+	existingDelegation, exists := td.mapping[stakingTxHash]
+	if !exists {
+		// If it doesn't exist, return false for changed, and false for exists
+		return false, false
+	}
+
+	// Compare fields to check if the delegation has changed
+	if existingDelegation.StakingOutputIdx != newDelegation.stakingOutputIdx ||
+		!reflect.DeepEqual(existingDelegation.UnbondingOutput, newDelegation.unbondingOutput) ||
```
KonradStaniec, line 116: hmm I think the only thing that could have changes is `.DelegationStartHeight` ? I.e 
- `.StakingOutputIdx` will always be the same
- `.UnbondingOutput` will always be the same
-  `.DelegationStartHeight ` can change from `0` to some other value when providing proof of inclusion

So maybe lets be explicit here at compare only `DelegationStartHeight` ?

Lazar955: sg will amend

KonradStaniec (approved):

PR #75, opened 2024-10-10
ci: GH actions for goreleaser, changelog reminder, and backport

SebastianElvis:
Also bumped .github dependency


.github/workflows/goreleaser.yml
```diff
@@ -0,0 +1,11 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  release:
+    uses: babylonlabs-io/.github/.github/workflows/reusable_go_releaser.yml@v0.7.0
```
Lazar955, line 10: Should each repo owner add .goreleaser.yml to the root  of the project? How does the gh action know to build our binaries otherwise?

Lazar955 (approved):
Nice work? Some questions:

.github/workflows/goreleaser.yml
SebastianElvis: Hmm good question. This workflow does not upload binaries at the moment. To this end one needs to add stuff like https://github.com/marketplace/actions/goreleaser-action#upload-artifacts to the template. Can do in subsequent PRs

Lazar955: But I think it's also not building them either (how do we check it?), it needs something like [this](https://github.com/osmosis-labs/osmosis/blob/main/.goreleaser.yaml)

SebastianElvis: https://github.com/babylonlabs-io/pm/issues/78

SebastianElvis: > But I think it's also not building them either (how do we check it?), it needs something like [this](https://github.com/osmosis-labs/osmosis/blob/main/.goreleaser.yaml?rgh-link-date=2024-10-10T07%3A00%3A27Z)

Yeah the current workflow is just creating a release corresponding to the tagged version. To fix this one needs to first call `make build`, and then add a step in the workflow like [this](https://github.com/marketplace/actions/goreleaser-action#upload-artifacts) to upload the binary. Need some research there

PR #79, opened 2024-10-14
fix(submitter): handle no change output

Lazar955:
Handles a case where there are no change outputs

### DIFF
--- a/submitter/relayer/relayer.go
+++ b/submitter/relayer/relayer.go
@@ -487,7 +487,9 @@ func (rl *Relayer) ChainTwoTxAndSend(data1 []byte, data2 []byte) (*types.BtcTxIn
 func (rl *Relayer) buildTxWithData(data []byte, firstTx *wire.MsgTx) (*types.BtcTxInfo, error) {
 	tx := wire.NewMsgTx(wire.TxVersion)
 
-	if firstTx != nil {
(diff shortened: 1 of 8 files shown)


types/ckpt_info.go
```diff
@@ -18,9 +18,8 @@ type CheckpointInfo struct {
 
 // BtcTxInfo stores information of a BTC tx as part of a checkpoint
 type BtcTxInfo struct {
-	TxId          *chainhash.Hash
-	Tx            *wire.MsgTx
-	ChangeAddress btcutil.Address
```
Lazar955, line 23: We're not using `ChangeAddress`, removed it

submitter/relayer/relayer.go
```diff
@@ -487,7 +487,9 @@ func (rl *Relayer) ChainTwoTxAndSend(data1 []byte, data2 []byte) (*types.BtcTxIn
 func (rl *Relayer) buildTxWithData(data []byte, firstTx *wire.MsgTx) (*types.BtcTxInfo, error) {
 	tx := wire.NewMsgTx(wire.TxVersion)
 
-	if firstTx != nil {
+	isFirstTx := firstTx != nil
```
KonradStaniec, line 490: should this case be called is `isSecondTx` ? i.e if the first tx is not nil then we are building second tx?

PR #84, opened 2024-10-18
fix(sew): limit go routine spawn for activating delegation

Lazar955:
Limit go routines getting spawned and doing extra work.

### DIFF
--- a/btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
+++ b/btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
@@ -177,7 +177,7 @@ func (sew *StakingEventWatcher) checkBabylonDelegations(status btcstakingtypes.B
 			return fmt.Errorf("error fetching active delegations from babylon: %v", err)
 		}
 
-		sew.logger.Debugf("fetched %d delegations from babylon", len(delegations))
+		sew.logger.Debugf("fetched %d delegations from babylon by status %s", len(delegations), status)
 
 		for _, delegation := range delegations {
 			addDel(delegation)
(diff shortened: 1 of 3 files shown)


btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
```diff
@@ -539,6 +539,10 @@ func (sew *StakingEventWatcher) checkBtcForStakingTx() error {
 	}
 
 	for _, del := range delegations {
+		if del.ActivationInProgress {
```
RafilxTenfen, line 542: Would be better if we didn't get all the tracked delegations at once, because it could happen that we load the delegations to iterate, after that someone updates the `del.ActivationInProgress` value and this iterator would not know about it 

If we are okay with this scenario the code looks good to me

Lazar955: I do agree with what you say. Tho I do think there's not a high chance for this to happen. What approach would you recommend, first thing that come to my mind was using Iterators

RafilxTenfen: yes, like an iterator would work or a function which gets the next del that is not activated

KonradStaniec (approved):

btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
Lazar955: will create an issue to improve this, not a high priority, worst case is we retry

PR #87, opened 2024-10-23
feat(sew): ADR-029 generalized unbonding and babylon v0.13.0

Lazar955:
Extends staking event watcher to be able to send whole stake spending tx as well as inclusion proof.

### DIFF
--- a/btcstaking-tracker/stakingeventwatcher/expected_babylon_client.go
+++ b/btcstaking-tracker/stakingeventwatcher/expected_babylon_client.go
@@ -1,6 +1,7 @@
 package stakingeventwatcher
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 
@@ -10,7 +11,6 @@ import (
 	bbnclient "github.com/babylonlabs-io/babylon/client/client"
 	bbn "github.com/babylonlabs-io/babylon/types"
 	btcstakingtypes "github.com/babylonlabs-io/babylon/x/btcstaking/types"
-	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/cosmos/cosmos-sdk/types/query"
@@ -28,7 +28,8 @@ type BabylonNodeAdapter interface {
 	DelegationsByStatus(status btcstakingtypes.BTCDelegationStatus, offset uint64, limit uint64) ([]Delegation, error)
 	IsDelegationActive(stakingTxHash chainhash.Hash) (bool, error)
 	IsDelegationVerified(stakingTxHash chainhash.Hash) (bool, error)
-	ReportUnbonding(ctx context.Context, stakingTxHash chainhash.Hash, stakerUnbondingSig *schnorr.Signature) error
+	ReportUnbonding(ctx context.Context, stakingTxHash chainhash.Hash, stakeSpendingTx *wire.MsgTx,
+		inclusionProof *btcstakingtypes.InclusionProof) error
 	BtcClientTipHeight() (uint32, error)
 	ActivateDelegation(ctx context.Context, stakingTxHash chainhash.Hash, proof *btcctypes.BTCSpvProof) error
 }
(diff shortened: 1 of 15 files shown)


btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
```diff
@@ -426,7 +427,7 @@ func (sew *StakingEventWatcher) watchForSpend(spendEvent *notifier.SpendEvent, t
 		return
 	}
 
-	schnorrSignature, err := tryParseStakerSignatureFromSpentTx(spendingTx, td)
+	_, err := tryParseStakerSignatureFromSpentTx(spendingTx, td)
```
KonradStaniec, line 430: this function actually checks whether stake spending transaction is unbonding transaction registered to Babylon. 

So imo we can leave it but we should send inclusion proof through `sew.reportUnbondingToBabylon(quitCtx, delegationId, spendingTx, proof)` even if it returns error (as now we have possibility to report even invalid unbondings to Babylon)

So the only difference between error case and non err case is metrics which we are incrementing.

btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
```diff
@@ -454,6 +461,65 @@ func (sew *StakingEventWatcher) watchForSpend(spendEvent *notifier.SpendEvent, t
...
+	txHash := spendingTx.TxHash()
+	if len(spendingTx.TxOut) == 0 {
+		return nil, fmt.Errorf("stake spending tx has no outputs")
+	}
+	details, status, err := sew.btcClient.TxDetails(&txHash, spendingTx.TxOut[0].PkScript)
+	if err != nil {
+		return nil, err
+	}
+
+	if status != btcclient.TxInChain {
```
KonradStaniec, line 474: I wonder whether we need this check i.e whether ` <-spendEvent.Spend:` return when tx is in the mempool or in chain. (eitherway better to have it just in case)

btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
```diff
@@ -454,6 +461,65 @@ func (sew *StakingEventWatcher) watchForSpend(spendEvent *notifier.SpendEvent, t
 	)
 }
 
+func (sew *StakingEventWatcher) buildSpendingTxProof(spendingTx *wire.MsgTx) (*btcstakingtypes.InclusionProof, error) {
+	txHash := spendingTx.TxHash()
+	if len(spendingTx.TxOut) == 0 {
+		return nil, fmt.Errorf("stake spending tx has no outputs")
```
KonradStaniec, line 467: tbh this should not happen as we are handling tx after we detected it was included in mempool/chain. Maybe panic here or have some error which we do not retry ? (imo this means our assumptions are broken so we should do something drastic)

btcstaking-tracker/stakingeventwatcher/expected_babylon_client.go
```diff
@@ -111,13 +112,20 @@ func (bca *BabylonClientAdapter) IsDelegationVerified(stakingTxHash chainhash.Ha
 func (bca *BabylonClientAdapter) ReportUnbonding(
 	ctx context.Context,
 	stakingTxHash chainhash.Hash,
-	stakerUnbondingSig *schnorr.Signature) error {
+	stakeSpendingTx *wire.MsgTx,
+	inclusionProof *btcstakingtypes.InclusionProof) error {
 	signer := bca.babylonClient.MustGetAddr()
 
+	var stakeSpendingTxBuf bytes.Buffer
+	if err := stakeSpendingTx.Serialize(&stakeSpendingTxBuf); err != nil {
```
KonradStaniec, line 120: on `bbn` (`bbn "github.com/babylonlabs-io/babylon/types"`)we have this helper - `bbn.SerializeBTCTx(`

btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
```diff
@@ -437,13 +438,24 @@ func (sew *StakingEventWatcher) watchForSpend(spendEvent *notifier.SpendEvent, t
 		// As we only care about unbonding transactions, we do not need to take additional actions.
 		// We start polling babylon for delegation to stop being active, and then delete it from unbondingTracker.
 		sew.logger.Debugf("Spending tx %s for staking tx %s is not unbonding tx. Info: %v", spendingTxHash, delegationId, err)
+		proof, err := sew.waitForStakeSpendInclusionProof(quitCtx, spendingTx)
+		if err != nil {
+			sew.logger.Errorf("unbonding tx %s for staking tx %s proof not built", spendingTxHash, delegationId)
+			return
+		}
+		sew.reportUnbondingToBabylon(quitCtx, delegationId, spendingTx, proof)
 		sew.waitForDelegationToStopBeingActive(quitCtx, delegationId)
```
KonradStaniec, line 447: we probably do not need it here any more, as after succesful unbonding report delegation won't be active (and `reportUnbondingToBabylon` has acitvity check inside)

KonradStaniec (approved):

PR #98, opened 2024-11-09
chore(lint): fix golangci config and linting errors

Lazar955:
Fixes the format of the golangci configuration and lining errors. Previously when using golangci-lint it didn't read the configurations so we only had default checks

### DIFF
--- a/btcclient/query.go
+++ b/btcclient/query.go
@@ -82,6 +82,7 @@ func (c *Client) getBestBlockHashWithRetry() (*chainhash.Hash, error) {
 	); err != nil {
 		c.logger.Debug(
 			"failed to query the best block hash", zap.Error(err))
+		return nil, err
 	}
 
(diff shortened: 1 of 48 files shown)


btcclient/query.go
```diff
@@ -82,6 +82,7 @@ func (c *Client) getBestBlockHashWithRetry() (*chainhash.Hash, error) {
 	); err != nil {
 		c.logger.Debug(
 			"failed to query the best block hash", zap.Error(err))
+		return nil, err
```
KonradStaniec, line 85: woh those seems like critical fixes 😅

KonradStaniec (approved):

btcclient/query.go
Lazar955: yeah some obvious misses on our end 😅

PR #102, opened 2024-11-14
chore(*): upgrade babylon to v0.16.0

Lazar955:
Upgrades babylon


go.mod
```diff
@@ -1,24 +1,26 @@
 module github.com/babylonlabs-io/vigilante
 
-go 1.23
+go 1.23.1
+
+toolchain go1.23.3
```
gitferry, line 5: just curious, why go version and tool chain does not match?

gitferry (approved):

PR #120, opened 2024-11-27
chore(e2e): test multiple unbonding txns

Lazar955:
# ⚠️ ⚠️ ⚠️  DO NOT MERGE ⚠️ ⚠️ ⚠️ 

Showcase that we can support the processing of 1000 unbonding transactions in a single block. At the end of the test, we measure the time taken to wait for all of them to be unbonded.

Test will take a lot of time, as we need to submit delegations.

[References](https://github.com/babylonlabs-io/pm/issues/131)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ test:
 	go test -race ./...
 
 test-e2e:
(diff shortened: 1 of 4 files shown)


Lazar955:
This was a successful benchmark, closing the PR

PR #127, opened 2024-11-29
chore(sew): improvments to staking event tracker

Lazar955:
This should solve a long lock wait time, which happens when fetching delegations into the tracker and trying to update a tracker

### DIFF
--- a/btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
+++ b/btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
@@ -75,6 +75,7 @@ type StakingEventWatcher struct {
 	babylonNodeAdapter BabylonNodeAdapter
 	unbondingTracker   *TrackedDelegations
 	pendingTracker     *TrackedDelegations
+	inProgressTracker  *TrackedDelegations
 
 	unbondingDelegationChan chan *newDelegation
 
@@ -100,6 +101,7 @@ func NewStakingEventWatcher(
 		metrics:                 metrics,
 		unbondingTracker:        NewTrackedDelegations(),
 		pendingTracker:          NewTrackedDelegations(),
+		inProgressTracker:       NewTrackedDelegations(),
 		unbondingDelegationChan: make(chan *newDelegation),
 		unbondingRemovalChan:    make(chan *delegationInactive),
 	}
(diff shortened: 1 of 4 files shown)


btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
```diff
@@ -100,6 +101,7 @@ func NewStakingEventWatcher(
 		metrics:                 metrics,
 		unbondingTracker:        NewTrackedDelegations(),
 		pendingTracker:          NewTrackedDelegations(),
+		inProgressTracker:       NewTrackedDelegations(),
```
KonradStaniec, line 104: Some doc comment would great about what exactly is `inProgressTracker` delegation ?

btcstaking-tracker/stakingeventwatcher/tracked_delegations.go
```diff
@@ -105,6 +105,28 @@ func (td *TrackedDelegations) AddDelegation(
 	return delegation, nil
 }
 
+func (td *TrackedDelegations) AddDel(
```
KonradStaniec, line 108: hmm we now have to function to add to tracker `AddDelegation` and `AddDel` ? How should the caller know which function to use ? It would be great to:
- either have some docs about when each methods should be used
- or have only one methods to add to trackers

KonradStaniec (approved):
2 non-blocking comments.

Tbh it is getting a bit hard to follow what is happening inside the event tracker, and when each of the `TrackedDelegations` objects  should be used.

btcstaking-tracker/stakingeventwatcher/tracked_delegations.go
Lazar955: Good point, not sure where my head was, removed the AddDel, it's completely unnecessary

btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
Lazar955: Agreed, added a doc string. 

inProgressTracker is basically a cache that says, hey we are trying to activate this (in a go routine) so let's not try again, hence the check in the for loop, and removal in the go routine from it.

PR #139, opened 2024-12-11
feat: add opcc slashing event

parketh:
## Summary

This is a PR to update the BTC Slasher program to support and handle OP stack consumer chain slashing events. 

It accompanies https://github.com/babylonlabs-io/babylon-contract/pull/92.

## Test plan

Before running integration tests, I needed to set `DOCKER_HOST`:
```bash
export DOCKER_HOST="unix:///Users/parkyeung/.docker/run/docker.sock"
```

To run tests:

```bash
# unit tests
make test

# integration tests
make test-e2e
```


CHANGELOG.md
```diff
@@ -37,6 +37,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+* [#135](https://github.com/babylonlabs-io/vigilante/pull/135) add opcc slashing event
+
```
lesterli, line 41: it should be #139

lesterli (approved):

btcstaking-tracker/btcslasher/slasher.go
```diff
@@ -132,10 +135,18 @@ func (bs *BTCSlasher) Start() error {
...
 		bs.finalitySigChan, startErr = bs.BBNQuerier.Subscribe(txSubscriberName, queryName)
 		if startErr != nil {
 			return
 		}
+		// subscribe to consumer fp slashing events
+		queryName = fmt.Sprintf("tm.event = 'Tx' AND message.action='%s'", consumerMessageActionName)
+		bs.equivocationEvidenceChan, startErr = bs.BBNQuerier.Subscribe(txSubscriberName, queryName)
+		if startErr != nil {
+			return
+		}
```
SebastianElvis, line 148: Probably one can use a OR statement to subscribe to both in a single Subscribe routine

SebastianElvis (approved):
Thanks!

@KonradStaniec Is it okay that for the `main` branch we have this piece of code for integration?

Lazar955 (approved):

btcstaking-tracker/btcslasher/slasher.go
parketh: tried doing it but it wasnt that straightforward as we then need to parse and sort the incoming messages. the current implementation seems a bit cleaner overall

SebastianElvis: sounds good

KonradStaniec (approved):
> @KonradStaniec Is it okay that for the main branch we have this piece of code for integration?

This seems like small piece of code so it should be fine 👍 

Also e2e tests are passing so it should it seems that it is possible to subscribe to the messages that do not exists yet on Babylon (`MsgEquivocationEvidence`)

PR #166, opened 2025-01-09
chore: refactor checkpoint tx submission

### LINKED ISSUE #163: Submitter: FundRawTransaction should choose appropriate utxo
Err we got:

> 2025-01-09T00:54:23.279457Z	error	Failed to submit the raw checkpoint for 12: failed to add data to tx: the value of the utxo is not sufficient for paying the calculated fee of the tx. Calculated: 0.00052920 BTC. Have: 0.00050244 BTC	{"module": "submitter"}

Lazar955:
Closes #163

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+### Improvements
+
+* [#166](https://github.com/babylonlabs-io/vigilante/pull/166) chore: refactor checkpoint tx submission
+
 ## v0.19.2
 
 ### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,9 +37,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
-### Improvements
+### Bug Fixes
 
-* [#166](https://github.com/babylonlabs-io/vigilante/pull/166) chore: refactor checkpoint tx submission
+* [#166](https://github.com/babylonlabs-io/vigilante/pull/166) fix: refactor checkpoint tx submission,
+fix change addr creation
 
 ## v0.19.2
 
--- a/submitter/relayer/relayer.go
+++ b/submitter/relayer/relayer.go
@@ -519,7 +519,7 @@ func (rl *Relayer) buildDataTx(data []byte) (*types.BtcTxInfo, error) {
 	hasChange := len(rawTxResult.Transaction.TxOut) > changePosition
 	if !hasChange {
 		rl.logger.Debugf("no change, adding change address manually, tx ref: %s", tx.TxHash())
-		changeAddr, err := rl.BTCWallet.GetRawChangeAddress("bech32")
+		changeAddr, err := rl.BTCWallet.GetNewAddress("")
 		if err != nil {
 			return nil, fmt.Errorf("err getting raw change address %w", err)
 		}
--- a/submitter/relayer/relayer.go
+++ b/submitter/relayer/relayer.go
@@ -518,7 +518,8 @@ func (rl *Relayer) buildDataTx(data []byte) (*types.BtcTxInfo, error) {
 	// Ensure there's a change output with at least dust amount
 	hasChange := len(rawTxResult.Transaction.TxOut) > changePosition
 	if !hasChange {
-		changeAddr, err := rl.BTCWallet.GetRawChangeAddress(rl.walletName)
+		rl.logger.Debugf("no change, adding change address manually, tx ref: %s", tx.TxHash())
+		changeAddr, err := rl.BTCWallet.GetRawChangeAddress("bech32")
 		if err != nil {
 			return nil, fmt.Errorf("err getting raw change address %w", err)
 		}
--- a/submitter/relayer/relayer.go
+++ b/submitter/relayer/relayer.go
@@ -430,7 +430,9 @@ func (rl *Relayer) retrySendTx2(ckpt *ckpttypes.RawCheckpointResponse) (*types.C
 		return nil, fmt.Errorf("tx1 is nil") // shouldn't happen, sanity check
 	}
 
-	tx2, err := rl.buildAndSendTx(data2, tx1.Tx)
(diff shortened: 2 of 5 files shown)


submitter/relayer/relayer.go
```diff
@@ -544,74 +532,105 @@ func (rl *Relayer) buildTxWithData(data []byte, firstTx *wire.MsgTx) (*types.Btc
...
-	}
-
-	txFee := rawTxResult.Fee
-	// ensuring the tx fee is not lower than the minimum relay fee
-	if txFee < minRelayFee {
-		txFee = minRelayFee
-	}
-	// ensuring the tx fee is not higher than the utxo value
+	txFee := rl.calcMinRelayFee(txSize)
 	if hasChange && changeAmount < txFee {
```
Lazar955, line 608: this is redundant, as we should rely on fundRawTx to estimate the fee correctly and pick the right utxo

submitter/relayer/relayer.go
```diff
@@ -475,62 +479,47 @@ func (rl *Relayer) ChainTwoTxAndSend(data1 []byte, data2 []byte) (*types.BtcTxIn
...
 
-	// we want to ensure that firstTx has change output, but for the second transaction we can ignore this
+	// Ensure there's a change output with at least dust amount
 	hasChange := len(rawTxResult.Transaction.TxOut) > changePosition
-	// let's manually add a change output with 546 satoshis
-	if !isSecondTx && !hasChange {
-		changeAddr, err := rl.BTCWallet.GetRawChangeAddress(rl.walletName)
+	if !hasChange {
+		rl.logger.Debugf("no change, adding change address manually, tx ref: %s", tx.TxHash())
+		changeAddr, err := rl.BTCWallet.GetRawChangeAddress("bech32")
```
Lazar955, line 522: note here, this was a bug, GetRawChangeAddress expects address type, not wallet

KonradStaniec: actually I think this should be wallet name i.e if you dig into btcd library it shows:
```
func (c *Client) GetRawChangeAddress(account string) (btcutil.Address, error) {
	return c.GetRawChangeAddressAsync(account).Receive()
}

func (c *Client) GetRawChangeAddressAsync(account string) FutureGetRawChangeAddressResult {
	cmd := btcjson.NewGetRawChangeAddressCmd(&account, nil)
	result := FutureGetRawChangeAddressResult{
		network:         c.chainParams,
		responseChannel: c.SendCmd(cmd),
	}
	return result
}

func NewGetRawChangeAddressCmd(account, addrType *string) *GetRawChangeAddressCmd {
	return &GetRawChangeAddressCmd{
		Account:     account,
		AddressType: addrType,
	}
}

```

so there isn't a way to pass `AddressType`. 

Maybe we should swtich here to use https://developer.bitcoin.org/reference/rpc/getnewaddress.html, I have never encountered problems with this one.

submitter/relayer/relayer.go
```diff
@@ -475,62 +479,47 @@ func (rl *Relayer) ChainTwoTxAndSend(data1 []byte, data2 []byte) (*types.BtcTxIn
...
 	dataScript, err := builder.AddOp(txscript.OP_RETURN).AddData(data).Script()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to build data script: %w", err)
 	}
 	tx.AddTxOut(wire.NewTxOut(0, dataScript))
 
-	changePosition := 1 // must declare here as you cannot take address of const needed bellow
+	// Fund the transaction
+	changePosition := 1
```
gitferry, line 508: can we use the const variable which is already defined?

KonradStaniec (commented):
refactoring looks good, I am not sure about the bug fix though 🤔

submitter/relayer/relayer.go
```diff
@@ -544,74 +533,105 @@ func (rl *Relayer) buildTxWithData(data []byte, firstTx *wire.MsgTx) (*types.Btc
...
+	// Build txOut for data
+	builder := txscript.NewScriptBuilder()
+	dataScript, err := builder.AddOp(txscript.OP_RETURN).AddData(data).Script()
+	if err != nil {
+		return nil, fmt.Errorf("failed to build data script: %w", err)
+	}
+	tx.AddTxOut(wire.NewTxOut(0, dataScript))
+
+	// Fund the transaction
+	changePosition := 1
```
gitferry, line 567: given that we already know it's the second tx of the checkpoint. Do we still care change output?

gitferry (approved):
Nice fix! Some minor comments:

submitter/relayer/relayer.go
Lazar955: we need a reference to pass to FundRawTransaction, that's why the redeclare as var, can't take a ref of const

submitter/relayer/relayer.go
Lazar955: @KonradStaniec  Which version of btcd lib is this? Because if I pass anything else then what's defined [here](https://bitcoincore.org/en/doc/27.0.0/rpc/wallet/getrawchangeaddress/) we get an error saying: -5: Unknown address type 'default'. (default being wallet name)

submitter/relayer/relayer.go
Lazar955: good question, @KonradStaniec wdyt?

CHANGELOG.md
```diff
@@ -37,6 +37,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+### Improvements
+
+* [#166](https://github.com/babylonlabs-io/vigilante/pull/166) chore: refactor checkpoint tx submission
```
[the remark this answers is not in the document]
Lazar955: actually yes :D

submitter/relayer/relayer.go
Lazar955: From testnet log: 
```
Failed to submit the raw checkpoint for 34: failed to add data to tx: err getting raw change address -5: Unknown address type 'signet'	{"module": "submitter"}
```

KonradStaniec (approved):

submitter/relayer/relayer.go
KonradStaniec: we do not care about change output of the second transaction but I would leave this enforcement here tbh. It will make all our op_returns to be at index `0` which will make them easy to find in tx.

PR #171, opened 2025-01-14
chore: adds more logs to staking event watcher

Lazar955:
Debug logs for testnet


KonradStaniec (approved):
Could you write in a bit more detail what was happning and how the changes fixes the problem (also it would be nice to have some test showing what happened can be done in separate pr)

Lazar955:
> Could you write in a bit more detail what was happning and how the changes fixes the problem (also it would be nice to have some test showing what happened can be done in separate pr)

Yes, created an issue for the testcase for this. 

The issue was a combination of high contention as there are write to the pending tracker and reads from it and the fact that we could be yielding the same item (due the fact that we are writing to a map), fixed by the `processed` map. We were never stopping with processing, but processing them really really slow

PR #179, opened 2025-01-15
chore: metrics for bs tracker

### LINKED ISSUE #172: staking event watcher extra metrics
- count of verified but not in the chain
- count of verified but in the chain (not yet sufficient num of confirmations)
- count of verified in the chain with sufficient confirmations

Lazar955:
Adds 2 new metrics for bs tracker

closes #172

### DIFF
--- a/btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
+++ b/btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
@@ -82,6 +82,11 @@ type StakingEventWatcher struct {
 	// keeps track of delegations that are verified, and we are trying to activate
 	inProgressTracker *TrackedDelegations
 
+	// used for metrics purposes, keeps track of verified delegations that are not in consumer chain
+	verifiedNotInChainTracker *TrackedDelegations
+	// used for metrics purposes, keeps track of verified delegations in chain but without sufficient confirmations
+	verifiedInsufficientConfTacker *TrackedDelegations
+
 	unbondingDelegationChan chan *newDelegation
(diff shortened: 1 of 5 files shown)


btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
```diff
@@ -97,19 +102,21 @@ func NewStakingEventWatcher(
...
+		logger:                         parentLogger.With(zap.String("module", "staking_event_watcher")).Sugar(),
+		btcNotifier:                    btcNotifier,
+		btcClient:                      btcClient,
+		babylonNodeAdapter:             babylonNodeAdapter,
+		metrics:                        metrics,
+		unbondingTracker:               NewTrackedDelegations(),
+		pendingTracker:                 NewTrackedDelegations(),
+		inProgressTracker:              NewTrackedDelegations(),
+		verifiedInsufficientConfTacker: NewTrackedDelegations(),
+		verifiedNotInChainTracker:      NewTrackedDelegations(),
```
RafilxTenfen, line 116: this seems quite expensive to keep in memory

RafilxTenfen: if we are okay with keeping track of all of this state I am good to approve

Lazar955: Those 2 extra maps are adding each 1mb per 10k entries, I think we should be fine

RafilxTenfen (approved):
utACK

PR #184, opened 2025-01-17
fix: removing from metrics tracker

Lazar955:
We were returning nil, making the statement never true


RafilxTenfen (approved):
good catch, lets also add a unit test?

Lazar955:
> good catch, lets also add a unit test?

will add in the next PR, added an e2e condition on the metric

PR #190, opened 2025-01-23
chore: adds more metrics

Lazar955:
Adds new metric for tracker, less log noise


RafilxTenfen (approved):
LGTM, there is anyway to unit test this metrics properties?

Lazar955:
> LGTM, there is anyway to unit test this metrics properties?

yep, we already are in e2e

PR #194, opened 2025-01-24
chore: reduce locks

Lazar955:


btcstaking-tracker/stakingeventwatcher/tracked_delegations.go
```diff
@@ -231,8 +236,5 @@ func (td *TrackedDelegations) UpdateActivation(tx chainhash.Hash, inProgress boo
 }
 
 func (td *TrackedDelegations) Count() int {
-	td.mu.RLock()
-	defer td.mu.RUnlock()
-
-	return len(td.mapping)
+	return int(td.count.Load())
```
Lazar955, line 239: We use count for metrics often, in order to reduce lock contention we introduce atomic counter

btcstaking-tracker/stakingeventwatcher/tracked_delegations.go
```diff
@@ -18,9 +18,10 @@ type TrackedDelegation struct {
 }
 
 type TrackedDelegations struct {
-	mu sync.RWMutex
+	mu deadlock.RWMutex
```
Lazar955, line 21: might remove this in future, wanted to see if we will catch any deadlocks

btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
```diff
@@ -620,11 +616,11 @@ func (sew *StakingEventWatcher) checkBtcForStakingTx() {
 	}
 
 	for del := range sew.pendingTracker.DelegationsIter(1000) {
-		if inProgDel, _ := sew.inProgressTracker.GetDelegation(del.StakingTx.TxHash()); inProgDel != nil {
+		if del.ActivationInProgress {
```
Lazar955, line 619: simplifies things, no need to have 2 trackers and potentially introduce deadlocks/contention

btcstaking-tracker/stakingeventwatcher/tracked_delegations.go
KonradStaniec: given that `count` is only updated when mutex `mu` is locked, maybe it makes sense:
- either have `count` as normal variable and mutex locked for reading
- or refactor code somehow to avoid updating atomic variable inside the mutex
i.e in my mind updating atomic variable inside the lock is kind of code smell which indicates something can be better. 

Though if we already determined that this approach reduces contention considerably, we can leave it as it is and investigate later.

KonradStaniec (approved):

btcstaking-tracker/stakingeventwatcher/tracked_delegations.go
Lazar955: I agree with you that it looks like a code smell, but this enables us to call `Count()` (in-hot path) while not obtaining a lock Having an `int` counter which would need to obtain a lock is similar to just calling `len` on the map

PR #195, opened 2025-01-27
chore: update bbn to rc4

Lazar955:


btcclient/client_wallet.go
```diff
@@ -46,7 +46,8 @@ func NewWallet(cfg *config.Config, parentLogger *zap.Logger) (*Client, error) {
 
 	connCfg := &rpcclient.ConnConfig{
 		// this will work with node loaded with multiple wallets
-		Host:         cfg.BTC.Endpoint + "/wallet/" + cfg.BTC.WalletName,
```
KonradStaniec, line 49: this seems as bit unrelated to the bump. Why it was needed ?

KonradStaniec (approved):

btcclient/client_wallet.go
Lazar955: Yep, now we get an error: `tcp/:port/wallet/:ourwalletName` cannot connect to a port or something like that. This bump bumps btcd lib also, and also bumps lnd as a consequence of that

PR #206, opened 2025-01-29
chore: perf test unbonding reporting

Lazar955:
WIP


Lazar955:
We are going with indexer rather than batching as it's more scalable long term, closing the PR

PR #210, opened 2025-02-07
feat: add indexer in e2e

Lazar955:
Keeping PRs small, this one just adds indexer container which is connected to the bitcoind node


RafilxTenfen (approved):
:rocket:

PR #223, opened 2025-02-11
fix: consider minimal fee for bump

### LINKED ISSUE #222: submitter: Wrong calculation of default transaction fee bump prevents bumping transactions
The relayer's Replace-by-Fee (RBF) mechanism for Bitcoin transactions will not work under the following scenarios:
1. When using the default ResubmitFeeMultiplier value
2. When the current fee requirements exceed the fee chosen for the Replace-by-Fee transaction.

Compute the replacement transaction fee considering the current network fee. 

**bp2-018**

Lazar955:
closes #222

bp2-018


KonradStaniec (approved):
general comment, we need to start to test al those weird bitcoin related rules.

PR #226, opened 2025-02-12
fix: reselect inputs after adding manual output

### LINKED ISSUE #216: Transaction execution at risk due to decreased feerate
Re-select transaction inputs after adding the change output.

In `vigilante/submitter/relayer/relayer.go` : 
```go
// Ensure the firstTx has a change output, but we can skip this for the
// second transaction
hasChange := len(rawTxResult.Transaction.TxOut) > changePosition
// Manually add a change output with 546 satoshis if needed
if !isSecondTx && !hasChange {
changeAddr, err := rl.BTCWallet.GetRawChangeAddress(rl.walletName)
if err != nil {
return nil, fmt.Errorf("error getting raw change address: %w",
err)
}
changePkScript, err := txscript.PayToAddrScript(changeAddr)
if err != nil {
return nil, fmt.Errorf("failed to create script for change
address: %s, error: %w", changeAddr, err)
}
changeOutput := wire.NewTxOut(int64(dustThreshold), changePkScript)
rawTxResult.Transaction.AddTxOut(changeOutput)
}

```

Lazar955:
closes #216 

BP2-015

### DIFF
--- a/submitter/relayer/relayer.go
+++ b/submitter/relayer/relayer.go
@@ -533,6 +533,17 @@ func (rl *Relayer) buildDataTx(data []byte) (*types.BtcTxInfo, error) {
 
 		changeOutput := wire.NewTxOut(int64(dustThreshold), changePkScript)
 		rawTxResult.Transaction.AddTxOut(changeOutput)
+
+		// reselect the inputs as the change output has been added
+		rawTxResult, err = rl.BTCWallet.FundRawTransaction(rawTxResult.Transaction, btcjson.FundRawTransactionOpts{
+			FeeRate:        &feeRate,
+			ChangePosition: &changePosition,
+		}, nil)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fund raw tx after nochange: %w", err)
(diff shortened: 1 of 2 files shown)


submitter/relayer/relayer.go
```diff
@@ -533,6 +533,15 @@ func (rl *Relayer) buildDataTx(data []byte) (*types.BtcTxInfo, error) {
 
 		changeOutput := wire.NewTxOut(int64(dustThreshold), changePkScript)
 		rawTxResult.Transaction.AddTxOut(changeOutput)
+
+		// reselect the inputs as the change output has been added
+		rawTxResult, err = rl.BTCWallet.FundRawTransaction(rawTxResult.Transaction, btcjson.FundRawTransactionOpts{
```
RafilxTenfen, line 538: can we add a comment why do we need to do that?

Lazar955: updated the comment, thanks for pointing out

KonradStaniec (approved):

PR #232, opened 2025-02-14
chore: comply to rbf rules

### LINKED ISSUE #218: Submitter: Replacement transactions might not be accepted in mempool
Replacement transactions might be rejected due to:
1. Non-standard transactions.
2. Failure to meet Replace-by-Fee (RBF) minimum feerate.

### Issue Details
In `resendSecondTxOfCheckpointToBTC`, if the required replacement fee exceeds the change output, the entire change is used as fees. This can result in a dust output, making the transaction non-standard.

Additionally, since fees are capped by the change output, the transaction may fail to meet RBF minimum feerate. There are also no validations ensuring that the replacement fee is greater than the sum of the original transactions or that it covers the bandwidth cost.

### Recommendation
1. Ensure replacement transactions do not produce dust outputs.
2. Validate that replacement transactions meet the minimum required feerate.
3. Implement additional checks to ensure compliance with mempool replacement policies.

**BP2-019**

Lazar955:
Makes sure that `verifyRBFRequirements` is following the RBF [rules](https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md)

closes #218 

**BP2-019**


KonradStaniec (approved):
we need to start adding some unit tests for all those weird btc rules 😅

PR #234, opened 2025-02-18
chore: slasher poll for evidence

### LINKED ISSUE #231: Slasher: missing slashing events
Currently, the vigilante slasher is relying on `tm.event = 'Tx' and message.action='/babylon.finality.v1.MsgAddFinalitySig'` for events. 
But could potentially miss slashin if someone uses a `cosmowasm` contract or any kind of wrapped message as a proxy to call `AddFinalitySig`, this event would never be emitted. Instead, the event is `"message.action":["/cosmwasm.wasm.v1.MsgExecuteContract"]`.

Instead of relying on events, we could poll the chain for evidence, calling periodically `ListEvidences`. Currently, we have the mechanism during the bootstrap procedure of the slasher. https://github.com/babylonlabs-io/vigilante/blob/0abb1edd94788fe4b8670ba2985853a5548d13eb/btcstaking-tracker/btcslasher/bootstrapping.go#L26-L66

**BP2-029**

Lazar955:
We resolve the issue of missing certain events by polling the chain for evidence, this is also a better approach as cometbft doesn't have event delivery guarantees 

Closes #231


e2etest/test_manager.go
```diff
@@ -205,7 +205,9 @@ func (tm *TestManager) RetrieveTransactionFromMempool(t *testing.T, hashes []*ch
 	var txs []*btcutil.Tx
 	for _, txHash := range hashes {
 		tx, err := tm.BTCClient.GetRawTransaction(txHash)
-		require.NoError(t, err)
+		if err != nil {
+			continue
+		}
```
SebastianElvis, line 210: wdyt we return error upon this? it's possible that some hash is missing and it's the caller's responsibility to retry

btcstaking-tracker/btcslasher/bootstrapping.go
```diff
@@ -53,24 +70,25 @@ func (bs *BTCSlasher) Bootstrap(startHeight uint64) error {
...
-			// on Bitcoin
+			// Attempt to slash the finality provider
 			if err := bs.SlashFinalityProvider(fpBTCSK); err != nil {
 				bs.logger.Errorf("failed to slash finality provider %s: %v", fpBTCPKHex, err)
 				accumulatedErrs = multierror.Append(accumulatedErrs, err)
 
 				continue
 			}
+
+			lastSlashedHeight = evidence.BlockHeight
```
SebastianElvis, line 81:
```suggestion
            if lastSlashedHeight < evidence.BlockHeight {
			    lastSlashedHeight = evidence.BlockHeight
            }
```
Not sure if this check is needed as I don't know whether the returned evidences are ordered

btcstaking-tracker/btcslasher/slasher.go
```diff
@@ -339,3 +257,32 @@ func (bs *BTCSlasher) Stop() error {
...
+
+				continue
+			}
+
+			// Only update height if we actually processed evidence
+			if lastSlashedHeight > 0 {
+				bs.mu.Lock()
+				bs.height = lastSlashedHeight + 1
+				bs.mu.Unlock()
+			}
```
SebastianElvis, line 281: Does it make sense to set `bs.height` to the current height of Babylon chain if `lastSlashedHeight == 0`?

SebastianElvis (approved):

KonradStaniec (approved):

PR #235, opened 2025-02-18
chore: use file based lock

Lazar955:
we often have flaky tests due to mutex not properly working in the e2e for allocating port.

Trying to fix it with a file-based lock.

err that we often see:
```sh
API error (500): driver failed programming external connectivity on endpoint bitcoind-TestSlasher_SlashingUnbonding (5208fb0daff063a1ae7474082c19156de8b38fe74163adac195016b465db3f89): Error starting userland proxy: listen tcp4 0.0.0.0:35448: bind: address already in use
```

### DIFF
--- a/testutil/port.go
+++ b/testutil/port.go
@@ -10,6 +10,7 @@ import (
 )
 
 // Directory for storing lock files
+const lockDir = "/tmp/testutil_ports"
 
 // AllocateUniquePort tries to find an available TCP port on localhost
 // by testing multiple random ports within a specified range.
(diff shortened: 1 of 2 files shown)


testutil/port.go
```diff
@@ -27,37 +24,40 @@ func AllocateUniquePort(t *testing.T) int {
 		portRange = 30000
 	)
 
+	lockDir := t.TempDir()
```
RafilxTenfen, line 27: humm, this temp dir will use the same dir in every test?

Lazar955: good catch, fixed

RafilxTenfen (approved):
It is worth trying, but I am suspicious why the mutex was not working properly

PR #236, opened 2025-02-19
fix: handle has inclusion proof err

### LINKED ISSUE #183: BTC staking tacker: handle activating delegation which has inclusion proof
> debug	retrying	{"attempt": 2, "max_attempts": 5, "error": "rpc error: code = Unknown desc = rpc error: code = Unknown desc = failed to execute message; message index: 0: the delegation d1a726bee4e62f753a5efe32820b063a9aeb22314d4b211c74585c2ecdccbce9 already has inclusion proof [cosmos/cosmos-sdk@v0.50.9/baseapp/baseapp.go:1023] with gas used: '50399': unknown request"}

Means that delagation has been activated, we should stop tracking it in the in progress and pending trackers

Lazar955:
closes #183


btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
```diff
@@ -762,6 +763,15 @@ func (sew *StakingEventWatcher) activateBtcDelegation(
 		if err := sew.babylonNodeAdapter.ActivateDelegation(ctx, stakingTxHash, proof); err != nil {
 			sew.metrics.FailedReportedActivateDelegations.Inc()
 
+			// someone already submitted inclusion proof, lets cleanup
+			if strings.Contains(err.Error(), "already has inclusion proof") {
```
KonradStaniec, line 767: it would be great to have e2e test for this as I am not sure whether error message are preserved here 🤔

Lazar955: It makes sense, added e2e with two trackers, and moved the cleanup logic, so we don't really care about the error, if the delegation is not in verified state, it needs to be cleaned up

KonradStaniec (approved):

PR #258, opened 2025-03-12
fix: Reject non-negative value and Zero for time Interval used by time.ticker

qj0r9j0vc2:
https://github.com/golang/go/blob/644b984027d11d43881507a938b28ed9df3b3320/src/time/tick.go#L37

in time.ticker, if checks if time.Interval is less than or equal with 0.

but vigilante doesn't check it and if I set to 0, it throws error.

screenshot


Lazar955 (approved):
Look good, thanks for your contribution! 🙌 
Just add the change in the change log under Unreleased > Bug fixes (see other PRs).

Lazar955 (changes requested):
changelog reminder

Lazar955 (approved):

PR #259, opened 2025-03-12
fix: calculate fee with rbf

Lazar955:


config/submitter.go
```diff
@@ -11,6 +11,9 @@ const (
 	DefaultPollingIntervalSeconds    = 60   // in seconds
 	DefaultResendIntervalSeconds     = 1800 // 30 minutes
 	DefaultResubmitFeeMultiplier     = 1
+	DefaultInsufficientFeeMargin     = 0.15
```
KonradStaniec, line 14: can we add some rationale why `0.15` is good default and how operators should chose those config values 
```
	InsufficientFeeMargin     float64 `mapstructure:"insufficient_fee_margin"`     // e.g. 0.15 for 15%
	InsufficientFeerateMargin float64 `mapstructure:"insufficient_feerate_margin"` // e.g. 0.15 for 15%
	FeeIncrementMargin        float64 `mapstructure:"fee_increment_margin"
```

i.e I am thinking of the case when somebody from devops team will try to change those value, then he should have good idea what control what

submitter/relayer/relayer.go
```diff
@@ -293,12 +308,85 @@ func (rl *Relayer) calculateBumpedFee(ckptInfo *types.CheckpointInfo) btcutil.Am
 	// Calculate the recommended fee using ResubmitFeeMultiplier
 	bumpedFee := ckptInfo.Tx2.Fee.MulF64(rl.config.ResubmitFeeMultiplier)
 
+	if previousFailure == nil ||
+		(!errors.Is(previousFailure, ErrInsufficientFee) &&
```
KonradStaniec, line 312: maybe we can move this three error checks to separate function with the name ? (it is a bit hard to read this thing 😅 )

submitter/relayer/relayer.go
```diff
@@ -293,12 +308,85 @@ func (rl *Relayer) calculateBumpedFee(ckptInfo *types.CheckpointInfo) btcutil.Am
...
+	mempoolEntry, err := rl.BTCWallet.GetMempoolEntry(txID)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get mempool entry for %s: %w", txID, err)
+	}
+
+	// Calculate adjustments based on error type and available mempool data
+	switch {
+	case errors.Is(previousFailure, ErrInsufficientFee):
+		// Get the total fees of original tx + descendants
+		originalTotalFees := btcutil.Amount(mempoolEntry.DescendantFees)
```
KonradStaniec, line 332: maybe each case can be separate function ie:
```go
case errors.Is(previousFailure, ErrInsufficientFee):
  fee  := handleInsufficientFee(..)
case errors.Is(previousFailure, ErrInsufficientFeerate):
  fee := handeInsufficientFeerate(..)
...
```

It will be much easier for the reader to distinguish the case at first glance, also each function can have a bit of doc how is the calculation fixes the error.

KonradStaniec (approved):
Some documentation / refactoring comments onlyh

PR #271, opened 2025-03-21
chore: support old and new genesis

Lazar955:


types/genesis_info.go
```diff
@@ -59,29 +60,54 @@ func GetGenesisInfoFromFile(filePath string) (*GenesisInfo, error) {
 	}
 
 	checkpointingGenState := checkpointingtypes.GetGenesisStateFromAppState(tmpBabylon.AppCodec(), appState)
-	err = checkpointingGenState.Validate()
-	if err != nil {
+	err1 := checkpointingGenState.Validate()
+	err2 := validateGenesisKeys(checkpointingGenState.GenesisKeys)
+	if err1 != nil && err2 != nil {
 		return nil, fmt.Errorf("invalid checkpointing genesis %w", err)
```
KonradStaniec, line 66: hmm probably need to return both errors (err1 and err2) as `err`  at this point will be about something different

types/pop.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"errors"
+	"github.com/babylonlabs-io/babylon/crypto/bls12381"
+	"github.com/babylonlabs-io/babylon/x/checkpointing/types"
+	"github.com/cometbft/cometbft/crypto/ed25519"
+	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
+)
+
+func validPop(blsPubkey bls12381.PublicKey, valPubkey cryptotypes.PubKey, ed25519Sig, blsSig []byte) bool {
```
KonradStaniec, line 11: maybe lets add some doc why we are doing this i.e in the TGE chain we have different format of PoP in genesis and we want to avoid migration and monitor check only genesis

KonradStaniec (approved):

types/genesis_info.go
Lazar955: ah good catch, typo here

PR #278, opened 2025-03-24
chore: metrics for censorship detection

### LINKED ISSUE #272: Metrics for censorship of in protocol messages
Every message vigiliante program send to Babylon is critical in protocol message. Vigiliante program should have the way of detecting that messages are send to the Babylon mempool but not included for long period of time. The amount of time should be configurable.

Initial idea is to set sufficiently large `BlockTimeout` config in the Babylon client, and in case of this error increment appropriate metrics.

- [ ] Implement in vigiliante
- [ ] Test on devnet

Lazar955:
Adds new metrics that indicate that censorship has happened. 

closes #272

### DIFF
--- a/metrics/btcstaking_tracker.go
+++ b/metrics/btcstaking_tracker.go
@@ -36,8 +36,8 @@ type UnbondingWatcherMetrics struct {
 	NumberOfVerifiedNotInChainDelegations       prometheus.Gauge
 	NumberOfVerifiedInsufficientConfDelegations prometheus.Gauge
 	NumberOfVerifiedSufficientConfDelegations   prometheus.Gauge
-
-	MethodExecutionLatency *prometheus.HistogramVec
+	MethodExecutionLatency                      *prometheus.HistogramVec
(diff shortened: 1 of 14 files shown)


metrics/btcstaking_tracker.go
```diff
@@ -111,6 +111,16 @@ func newUnbondingWatcherMetrics(registry *prometheus.Registry) *UnbondingWatcher
 			Name:      "unbonding_watcher_number_of_verified_sufficient_conf_delegations",
 			Help:      "The number of verified delegations with sufficient confirmations",
 		}),
+		UnbondingCensorshipGaugeVec: registerer.NewGaugeVec(
```
KonradStaniec, line 114: I would add also censorshop detection for sending inclusion proofs

KonradStaniec (commented):
Code wise looks good. Two things:
1. I would increase default block timeout to something like 10min in `DefaultBabylonConfig`, or in any place we configure `BlockTimeout
2. There is also `ReportSelectiveSlashing` method, which should track for censorship

KonradStaniec (approved):

PR #311, opened 2025-04-07
fix: add yaml metadata to cfg

Lazar955:


Lazar955:
doesn't handle default bbn config as it's imported from Babylon, closing

PR #318, opened 2025-04-10
fix: start metrics server before blocking reporter execution

qj0r9j0vc2:
### Summary

This PR addresses an issue where the Prometheus metrics server (`metrics.Start`) fails to start correctly when `vigilantReporter.Start()` is called first.

### Problem

In the previous implementation:

```go

			// start normal-case execution
			vigilantReporter.Start()

			// start Prometheus metrics server
			addr := fmt.Sprintf("%s:%d", cfg.Metrics.Host, cfg.Metrics.ServerPort)
			metrics.Start(addr, reporterMetrics.Registry)
			
``` 

### Solve

```go
			// start Prometheus metrics server
			addr := fmt.Sprintf("%s:%d", cfg.Metrics.Host, cfg.Metrics.ServerPort)
			metrics.Start(addr, reporterMetrics.Registry)


			// start normal-case execution
			vigilantReporter.Start()
			
```

if metric server starts after `vigilantReporter.Start()`, it metrics server doesn't start immediately.


Lazar955 (approved):
good catch, thanks!

Issue #322, opened 2025-04-15
Reporter bootstrap phase causes severe delay in checkpoint submission


qj0r9j0vc2:
# Summary
When the reporter component of Vigilante boots up, it performs a full re-processing of BTC blocks starting from height H - (k + w), which leads to a significant delay in actual checkpoint submission and the initialization of metrics.

This results in:

Poor responsiveness during startup

Unavailable metrics for monitoring

Drastically lower throughput compared to other reporters

Relevant Code
**bootstrapping.go**
https://github.com/babylonlabs-io/vigilante/blob/b1f3d435171c1cd0809f77ac64fbb4183dd19950/reporter/bootstrapping.go#L110-L120

**utils.go**
https://github.com/babylonlabs-io/vigilante/blob/49dbc76053aaee7b9ef41e963ec0035d6c46f17a/reporter/utils.go#L306-L324

https://github.com/babylonlabs-io/vigilante/blob/49dbc76053aaee7b9ef41e963ec0035d6c46f17a/reporter/utils.go#L222-L304

# Root Cause
The reporter performs a full traversal of the last k + w = 330 BTC blocks upon boot.

For each matched checkpoint, InsertBTCSpvProof() is called with retry logic:

Up to 5 retries, 400ms interval between retries

Including processing time, each checkpoint can take ~8–10 seconds

![Image](https://github.com/user-attachments/assets/5844f9dd-a1e7-40a7-a9fb-0ebef787f366)

Worst-case bootstrapping duration:
10s × 5 retries × 330 checkpoints ≈ ~4.5 hours

Metrics server only starts after bootstrap() completes, making it difficult to monitor startup progress.

Duplicate submission detection fails because the returned error is wrapped, making errors.Is() ineffective.

# Proposed Improvements
Run r.ProcessCheckpoints() on goroutine to avoid blocking startup and allow metrics to initialize earlier.

Implement custom submission logic in the reporter instead of relying on babylonClient.InsertBTCSpvProof():

This would allow direct access to the raw ErrDuplicatedSubmission error

Reduces overhead and latency in checkpoint verification

Alternatively, improve error unwrapping in the Babylon SDK to allow accurate error matching with errors.Is().

# Impact
Faster reporter startup time

Better system observability (metrics available earlier)

Higher checkpoint submission throughput

Reduced redundant processing during bootstrapping

Let me know if you'd like me to submit a draft PR or experiment with async bootstrapping. Happy to help! 🙌

[...]

PR #323, opened 2025-04-15
feat: improve reporter bootstrapping performance and reliability in checkpoint submission

qj0r9j0vc2:
# Summary
Improves the reporter’s bootstrapping performance by avoiding redundant checkpoint submissions and enabling non-blocking startup.

# Changes
Replaced InsertBTCSpvProof with ReliablySendMsg
→ Skips duplicated or finalized checkpoints using known error types

Wrapped ProcessCheckpoints() in a goroutine
→ Reporter no longer blocks startup
→ Metrics server starts immediately

Added type assertion for babylonClient
→ Ensures safe access to client methods

Preserved all metrics and logging behavior

# Result
Bootstrap time reduced from potentially hours to just a few seconds (if checkpoints already submitted)

# Context
We had been restarting the reporter frequently while tuning parameters, but bootstrapping never completed due to duplicated submission retries.
This affected metrics and responsiveness.
This PR addresses the root cause and improves operator experience.

# Related Issue
- https://github.com/babylonlabs-io/vigilante/issues/322


Lazar955 (approved):
Looks really good, thanks for your contribution. Fix lint err

PR #327, opened 2025-04-15
chore: bootstrap only once

Lazar955:
Make sure we don't have concurrent calls to bootstrap until we finish with async handling for the call to ProcessCheckpoints

### DIFF
--- a/reporter/bootstrapping.go
+++ b/reporter/bootstrapping.go
@@ -66,6 +66,28 @@ func (r *Reporter) bootstrap() error {
 		err                  error
 	)
 
+	r.bootstrapMutex.Lock()
+	if r.bootstrapInProgress {
+		r.bootstrapMutex.Unlock()
+		// we only allow one bootstrap process at a time
+		return fmt.Errorf("bootstrap already in progress")
+	}
+	r.bootstrapInProgress = true
+	r.bootstrapWg.Add(1)
+	r.bootstrapMutex.Unlock()
+
+	// flag to indicate if we should clean up, err happened
+	success := false
+	defer func() {
+		// cleanup func in case we error, prevents deadlocks
+		if !success {
+			r.bootstrapMutex.Lock()
+			r.bootstrapInProgress = false
+			r.bootstrapMutex.Unlock()
+			r.bootstrapWg.Done()
+		}
+	}()
+
 	// ensure BTC has caught up with BBN header chain
 	if err := r.waitUntilBTCSync(); err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


reporter/bootstrapping.go
```diff
@@ -147,6 +178,8 @@ func (r *Reporter) bootstrapWithRetries() {
 	ctx, cancel := r.reporterQuitCtx()
 	defer cancel()
 	if err := retry.Do(func() error {
+		r.bootstrapWg.Wait()
```
Lazar955, line 181: no need to retry if we are still not done with bootstrap

reporter/bootstrapping.go
```diff
@@ -66,6 +66,28 @@ func (r *Reporter) bootstrap() error {
...
+	defer func() {
+		// cleanup func in case we error, prevents deadlocks
+		if !success {
+			r.bootstrapMutex.Lock()
+			r.bootstrapInProgress = false
+			r.bootstrapMutex.Unlock()
+			r.bootstrapWg.Done()
+		}
+	}()
+
```
Lazar955, line 90: cleanup if we fail, make sure we can call bootstrap again

RafilxTenfen: This looks a bit weird to have this `!success` check, what do you think about migrate the bootstrap logic to another function and calling at the end 

and just have the defer as 

```go
	defer func() {
		r.setBootstrapInProgress(false)
                r.bootstrapWg.Done()
	}()
     return r.bootstrapLogic()
}

func (r *Reporter) setBootstrapInProgress(inProgress bool) {
	r.bootstrapMutex.Lock()
	r.bootstrapInProgress = inProgress
	r.bootstrapMutex.Unlock()
}
```

Lazar955: Why do you find it weird? Is it personal preference or an anti-pattern to have a cleanup like this?

RafilxTenfen: I believe it is personal preference due to having 2 defers that have the same code besides the success check

types/ckpt_cache.go
```diff
@@ -96,6 +96,8 @@ func (c *CheckpointCache) Match() {
 
 func (c *CheckpointCache) PopEarliestCheckpoint() *Ckpt {
```
KonradStaniec, line 97: hmm imo this function is still not thread safe as `HasCheckpoints ` lock is not related to newly introduced lock.

The following sequence of events is possible:
- `HasCheckpoints` atomically returns  `true` 
- but before new `c.mu.Lock()` is called, some go routine pop checkpoint
- `ckpt := c.Checkpoints[0]` is called on empty collection on panic

To make it thread safe, `HasCheckpoints` and poping the checkpoint must happen under same `lock`

reporter/bootstrapping.go
KonradStaniec, line 181: maybe lets document it why this `r.bootstrapWg.Wait()` is here ? I.e this is big foot gun for future maintainer that somehow `r.bootstrap()` increments `wg`, but waiting is here.

Lazar955: Agreed, added comments

types/ckpt_cache.go
Lazar955: Yep, good catch, amended in [159d6b2](https://github.com/babylonlabs-io/vigilante/pull/327/commits/159d6b2cc3a54fc8298d7a02a93df4cd4f0cff89)

PR #331, opened 2025-04-17
feat: adds send raw tx with burn amount

Lazar955:


btcstaking-tracker/btcslasher/slasher_utils.go
```diff
@@ -209,8 +209,9 @@ func (bs *BTCSlasher) sendSlashingTx(
 		fpBTCPK.MarshalHex(),
 	)
 
+	const maxBurnAmountBTC = float64(50000)
```
KonradStaniec, line 212: I would make move the constant to the top of the file. Also lets make sure api expects BTC not sats, maybe in e2e test ?

KonradStaniec (approved):
code looks good 👍 

Please just add e2e test showing that sending works as expected.

btcstaking-tracker/btcslasher/slasher_utils.go
Lazar955: added here [f81db1f](https://github.com/babylonlabs-io/vigilante/pull/331/commits/f81db1ffb3a73c79e878668e1959f6a256b57cec)

PR #356, opened 2025-05-26
chore(deps): bump babylon to v4

GAtom22:


Lazar955:
We might not want to merge this into main, because the main is compatible with current testnet/mainnet babylon

gusin13:
@Lazar955 in Babylon all the features are merged in main, then we have long living release branches release/v1.x, release/v2.x ..etc where relevant features are backported. 

Shouldn't we follow the same approach here? cc @SebastianElvis for vis

Lazar955:
Okay means we need a `release/v1.x`before merging this, because thus far vigilante was compatible with all node versions

PR #380, opened 2025-06-12
feat: stake expansion

RafilxTenfen:


GAtom22:
the e2e test are failing due to the non-existent docker img for the commit we're using. Once we merge the stake expansion changes on the release/v3 branch, I'llcreate a snapshot tag and use it here as dependency

CLAUDE.md
```diff
@@ -0,0 +1,78 @@
+# Vigilante - Babylon Bitcoin Bridge
```
Lazar955, line 1: Let's cleanup the AI slop. Too much unnecessary text :D

Lazar955 (commented):
Why did we bump bbn to v3?

CLAUDE.md
GAtom22: done

GAtom22:
> Why did we bump bbn to v3?

Good question. Is true that to just support the stake expansion feature is not strictly needed. However, for the e2e tests we need to bump the version to get the v3 features (stake expansion, signing ctx)

GAtom22:
tests failing on extracting priv key due to missing signing context in EvidenceResponse.
Addressed in https://github.com/babylonlabs-io/babylon/pull/1327

Lazar955 (approved):
LGTM! Tested against `8c11e8cd8b7ded2b25e252aa6a15840eaa349ffe` in e2e, means we are compatible with v2 node, so using v3 should be fine.

Lazar955 (approved):

PR #385, opened 2025-06-26
ci: publish image with workflow_dispatch event and bump workflow version

maiquanghiep:


RafilxTenfen:
I don't think it is related

PR #395, opened 2025-07-07
fix: remove mutex on pk in slasher

Lazar955:
Seems that we no longer get race condition on the PK. This was making slashing of delegations of a FP sequential

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+### Bug Fixes
+
+* [#395](https://github.com/babylonlabs-io/vigilante/pull/395) fix: remove mutex on pk in slasher
+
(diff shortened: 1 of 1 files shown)


btcstaking-tracker/btcslasher/slasher.go
```diff
@@ -204,9 +202,6 @@ func (bs *BTCSlasher) SlashFinalityProvider(extractedFpBTCSK *btcec.PrivateKey)
 		return fmt.Errorf("failed to get BTC delegations under finality provider %s: %w", fpBTCPK.MarshalHex(), err)
 	}
 
-	// Initialize a mutex protected *btcec.PrivateKey
-	safeExtractedFpBTCSK := types.NewPrivateKeyWithMutex(extractedFpBTCSK)
```
KonradStaniec, line 208: I wonder why the mutex was here ?

KonradStaniec (approved):
Looks good, just wonder why the mutex was here in the first place

PR #404, opened 2025-07-18
tryfix: patch vigilante

RafilxTenfen:
moved to babylon v3 fixed the parse of genesis wrapped create val msg


Lazar955 (changes requested):
Can you use the latest main to apply the fix? Not sure what branch you used for this, too much noise

PR #451, opened 2025-08-29
feat: stateful slasher

### LINKED ISSUE #436: Stateful slasher
We want Slasher to save last processed height, so that on boot we pickup were it left off.

Relevant issue https://github.com/babylonlabs-io/vigilante/issues/153

Lazar955:
closes #436


Lazar955:
cc @babylon-devops as this add additional config setting to the btc staking tracker

PR #455, opened 2025-09-05
chore: allow docker_publish to skip tests and bump workflow version

maiquanghiep:
Added input option to skip lint tests in workflow dispatch.


RafilxTenfen (approved):
utACK

PR #462, opened 2025-09-25
chore: add activation timing checker

samricotta:
- Adds a method that checks the timing of when a verified delegation becomes activated.
-

### DIFF
--- a/monitor/activation_unbonding_monitor.go
+++ b/monitor/activation_unbonding_monitor.go
@@ -175,6 +175,7 @@ func (m *ActivationUnbondingMonitor) processDelegations(verifiedDels []Delegatio
 			kDeep, err := m.CheckKDeepConfirmation(&del)
 			if err != nil {
 				m.logger.Warnf("Error checking K-deep for %s: %v", stakingTxHash, err)
+
 				return err
 			}
 
@@ -182,9 +183,7 @@ func (m *ActivationUnbondingMonitor) processDelegations(verifiedDels []Delegatio
 			if kDeep {
 				m.handleKDeepDel(stakingTxHash)
 			} else {
-				if _, exists := m.activationTracker[stakingTxHash]; exists {
-					delete(m.activationTracker, stakingTxHash)
-				}
+				delete(m.activationTracker, stakingTxHash)
 			}
 			m.mu.Unlock()
 


monitor/activation_unbonding_monitor.go
```diff
@@ -2,30 +2,172 @@ package monitor
...
+	"github.com/babylonlabs-io/vigilante/config"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"go.uber.org/zap"
+	"sync"
+	"time"
 )
 
+type ActivationTracking struct {
+	StakingHash    chainhash.Hash
+	FirstSeenKDeep time.Time
```
Lazar955, line 15: naming suggestion KDeepAt

monitor/activation_unbonding_monitor.go
```diff
@@ -2,30 +2,172 @@ package monitor
...
+	LastChecked    time.Time
+	HasAlerted     bool
+}
+
 type ActivationUnbondingMonitor struct {
-	babylonClient BabylonBTCStakingClient
+	babylonClient     BabylonBTCStakingClient
+	btcClient         btcclient.BTCClient
+	cfg               *config.BTCStakingTrackerConfig
```
Lazar955, line 23: we shouldn't be using other processes cfg, monitor has it's own config, extend it

monitor/activation_unbonding_monitor.go
```diff
@@ -2,30 +2,172 @@ package monitor
...
+			m.logger.Warnf("Error checking K-deep for %s: %v", stakingTxHash, err)
+			continue
+		}
+
+		if kDeep {
+			if tracker, exists := m.activationTracker[stakingTxHash]; exists {
+				// need to now start the timing
+				timeSinceKDeep := time.Now().Sub(tracker.FirstSeenKDeep)
+
+				if timeSinceKDeep > 30*time.Minute && !tracker.HasAlerted {
```
Lazar955, line 140: have timeout for activation as config value

monitor/activation_unbonding_monitor.go
```diff
@@ -2,30 +2,172 @@ package monitor
...
+	for _, verifiedDel := range verifiedDels {
+		stakingTxHash := verifiedDel.StakingTx.TxHash()
+		kDeep, err := m.CheckKDeepConfirmation(&verifiedDel)
+		if err != nil {
+			m.logger.Warnf("Error checking K-deep for %s: %v", stakingTxHash, err)
+			continue
+		}
+
+		if kDeep {
+			if tracker, exists := m.activationTracker[stakingTxHash]; exists {
```
Lazar955, line 136: to many nested if hard to read, extract in functions

monitor/activation_unbonding_monitor.go
```diff
@@ -2,30 +2,172 @@ package monitor
...
+		return false, err
+	}
+
+	confirmations := currentHeight - details.BlockHeight
+
+	return confirmations >= bbnDepth, nil
+}
+
+func (m *ActivationUnbondingMonitor) CheckActivationTiming() error {
+	verifiedDels, err := m.GetVerifiedDelegations()
```
Lazar955, line 119: maybe we can stream delegations. When we have 100k delegations it takes a lot of time to fetch them all, ±10minuts. That means for each time tick that we execute this function we will wait another 10m to check these

monitor/activation_unbonding_monitor.go
```diff
@@ -2,30 +2,172 @@ package monitor
...
+					HasAlerted:     false}
+			}
+		} else {
+			if _, exists := m.activationTracker[stakingTxHash]; exists {
+				delete(m.activationTracker, stakingTxHash)
+			}
+		}
+	}
+
+	for hash, tracker := range m.activationTracker {
```
Lazar955, line 159: this potentially can be it's own go routine, but it will intorduce complexity

Lazar955: I'd suggest batching in go routines, have 50 concurent go routines to fetch if you don't want this as a separate go routine. 10k verified delegatons, each request is around 50ms optimistically from RPC

monitor/activation_unbonding_monitor.go
```diff
@@ -1,35 +1,49 @@
...
+	HasAlerted  bool
+}
+
 type ActivationUnbondingMonitor struct {
-	babylonClient BabylonAdaptorClient
-	btcClient     btcclient.BTCClient
-	cfg           *config.BTCStakingTrackerConfig
+	babylonClient     BabylonAdaptorClient
+	btcClient         btcclient.BTCClient
+	Cfg               *config.MonitorConfig
```
Lazar955, line 26: no need to export cfg

monitor/activation_unbonding_monitor.go
```diff
@@ -88,3 +101,103 @@ func (m *ActivationUnbondingMonitor) CheckKDeepConfirmation(delegation *Delegati
...
+		}
+	}
+}
+
+func (m *ActivationUnbondingMonitor) processDelegations(verifiedDels []Delegation) {
+	var wg sync.WaitGroup
+	limiter := make(chan struct{}, 50)
+	for _, verifiedDel := range verifiedDels {
+		wg.Add(1)
+		limiter <- struct{}{}
```
Lazar955, line 162: Tell Claude to use `semaphore` and not proprietary channel to handle this. Also 50 is too small for limit, make it 1000

Lazar955 (approved):

PR #467, opened 2025-09-30
chore: prometheus metrics + wiring

samricotta:
- Add prom metrics in for activation and unbonding
- Wire up monitor service integration
- various wiring and updating of activationUnbondingTiming
- unit test for prom


Lazar955 (commented):
it would be nice to test the timeout, after all that's the main thing we are testing, you could define some mock with activation that take a long time, and in a config have shorter timeout.

Also it would be nice that when reporting the metric for a timeout, we also report the delegation ID in the metric (see how that is done) and log it, I would even use ERR level log for this

PR #471, opened 2025-10-07
chore: unbonding w e2e

samricotta:


e2etest/test_manager.go
```diff
@@ -241,6 +244,10 @@ func StartManager(t *testing.T, options ...TestManagerOption) *TestManager {
 		return true
 	}, eventuallyWaitTimeOut, eventuallyPollTime)
 
+	h := fnv.New64a()
+	h.Write([]byte(t.Name()))
+	seed := int64(h.Sum64())
```
Lazar955, line 249: why not just used timestamp for source of randomness?

e2etest/slasher_e2e_test.go
```diff
@@ -700,10 +706,9 @@ func TestSlasher_Loaded_MultiStaking(t *testing.T) {
 					stakingTrackerMetrics,
 					dbBackend,
 				)
-				go func() {
-					err := bsTracker.Bootstrap(0)
-					require.NoError(t, err)
-				}()
+				err = bsTracker.Bootstrap(0)
```
Lazar955, line 709: why is this no longer in go routine?

Lazar955 (commented):
looks good, just address the slasher e2e change

e2etest/slasher_e2e_test.go
Lazar955: yeah I think in that test we don;t want it to wait

PR #513, opened 2026-01-08
fix: replace panic with error return for non-unbonding spending paths

Lazar955:
## Summary
- Replace panic with error return when spending tx has < 4 witness elements
- Add `ErrSpendPathNotUnbonding` sentinel error for timelock path detection
- Skip Babylon reporting for timelock path spends (no unbonding to report)
- Fixes crash when processing expired delegations spent via timelock path

## Context

The `tryParseStakerSignatureFromSpentTx` function panicked when encountering a witness with fewer than 4 elements. This was a regression from the events tracking introduction - the old code only fetched ACTIVE delegations, but the new event-driven approach can encounter expired delegations being spent via the timelock path.

**Witness structures:**
- Timelock path: 3 elements (staker_sig, script, control_block)
- Unbonding path: 4+ elements (covenant_sigs, staker_sig, script, control_block)

## Changes

1. **Sentinel error**: Added `ErrSpendPathNotUnbonding` to identify timelock path spends
2. **Error wrapping**: `tryParseStakerSignatureFromSpentTx` now wraps the error with the sentinel
3. **Explicit handling**: `handleSpend` checks for this specific error and:
   - Skips `reportUnbondingToBabylon` (nothing to report for timelock spend)
   - Still removes delegation from tracker

This avoids unnecessary RPC calls to Babylon and makes the code flow explicit rather than relying on Babylon rejecting the report.

### DIFF
--- a/btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
+++ b/btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
@@ -357,6 +357,8 @@ func getStakingTxInputIdx(tx *wire.MsgTx, td *TrackedDelegation) (int, error) {
 
 // tryParseStakerSignatureFromSpentTx tries to parse staker signature from unbonding tx.
 // If provided tx is not unbonding tx it returns error.
+//
+//nolint:unparam // signature return may be used in future
 func tryParseStakerSignatureFromSpentTx(tx *wire.MsgTx, td *TrackedDelegation) (*schnorr.Signature, error) {
 	if len(tx.TxOut) != 1 {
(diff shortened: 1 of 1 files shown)


btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go
```diff
@@ -514,6 +520,21 @@ func (sew *StakingEventWatcher) handleSpend(ctx context.Context, spendingTx *wir
 		// We stop reporting if delegation is no longer active or we succeed.
 	}
 
+	// If spending tx is timelock path (not unbonding), skip reporting to Babylon
```
KonradStaniec: Maybe lets add comments that we are skipping report to Babylon becouse if timelock path is used it means that delegation already expired on Babylon side.

PR #541, opened 2026-04-15
fix(submitter): rehydrate lastSubmittedCheckpoint on restart

RafilxTenfen:
## Summary

Fixes a mainnet-impacting bug in the submitter where a restart while a
checkpoint is still unconfirmed in the BTC mempool leaves the in-memory
`rl.lastSubmittedCheckpoint` zero-valued, causing the RBF path to either
**panic** (v0.23.x) or **silently loop forever** (v0.24.x+).

## Context

Incident: mainnet PagerDuty alert *Checkpoint submission stalled* on the
submitter pod running **v0.23.10**.

Root cause (verified against the code, not just the alert):

1. `Relayer.New` (relayer.go:97) initialises `lastSubmittedCheckpoint`
   to `&types.CheckpointInfo{}` (Tx1=nil, Tx2=nil, Epoch=0, TS=0) and
   never rebuilds it from the persisted `SubmitterStore`.
2. On the first sealed checkpoint after restart, `SendCheckpointToBTC`
   runs `maybeResendFromStore`. Because both txs are still in the
   mempool it returns `hasBeenProcessed=true` and the function
   early-returns — **without touching `lastSubmittedCheckpoint`**.
3. `MaybeResubmitSecondCheckpointTx` then runs for the same epoch. `TS`
   is zero → `time.Since(TS)` is huge → the RBF retry loop enters.
4. Inside the retry, `calculateBumpedFee(rl.lastSubmittedCheckpoint, …)`
   runs against a nil `Tx2`:
   - **v0.23.10** (no #423 guard): nil deref at
     `ckptInfo.Tx2.Size` / `.Tx2.Fee` → **panic → crash loop**. This is
     what mainnet is hitting right now.
   - **v0.24.0-rc.3+** (includes #423 backport 6321a8c): returns
     `"checkpoint info or Tx2 is nil"`, retries 5×, increments
     `FailedResentCheckpointsCounter`, and **never makes progress** for
     that epoch. Quieter, still broken.

The #423 guard only converts a panic into a silent stall; it does not
let the daemon recover. Rehydration is the actual fix.

## Changes

- **`submitter/relayer/relayer.go`**
  - New `rehydrateLastSubmittedCheckpointFromStore(ckptEpoch)` that
    reconstructs `lastSubmittedCheckpoint` from the persisted
    `StoredCheckpoint`. Size is recomputed via the existing
    `calculateTxVirtualSize`; fee is pulled from the BTC node's mempool
    entry (`GetMempoolEntry`), with `calcMinRelayFee` as a fallback for
    the race where the tx was mined between `maybeResendFromStore` and
    here.
  - `SendCheckpointToBTC` calls the helper after
    `maybeResendFromStore` confirms the stored txs are in the mempool,
    so the subsequent RBF path always has valid `Tx1`/`Tx2`.
  - `MaybeResubmitSecondCheckpointTx` gains an early bail when
    `Tx1`/`Tx2` are still nil — defence in depth so we never retry
    against empty state even if a future path slips past the primary
    fix.
- **`submitter/relayer/relayer_test.go`**
  - `TestRehydrateLastSubmittedCheckpointFromStore`:
    - rehydrates from store using the mempool fee
    - falls back to min relay fee when the mempool entry is missing
    - no-op on epoch mismatch
    - no-op on empty store
  - `TestMaybeResubmitSecondCheckpointTx_EmptyState` verifies the early
    bail returns nil without entering the retry loop when state is
    empty.
- **`CHANGELOG.md`** — entry under *Unreleased / Bug Fixes*.

No schema changes, no migrations, no config changes.

## Deployment note

Mainnet is currently on **v0.23.10** which does not contain the #423
nil guard. That deploy still crashes on every restart while a ckpt sits
in the mempool. Options to unblock prod:

1. Backport this PR to `release/v0.23.x` and cut **v0.23.11**
   (recommended — drains the stuck checkpoint automatically on restart).
2. Upgrade mainnet to **v0.24.1** — safe from the submitter's
   perspective (no DB migration, only config-validation tightened from
   `< 0` to `<= 0` on a handful of interval fields) **but** v0.24.x
   requires a Babylon v4 node. Only an option if mainnet Babylon is
   already on v4. This PR is still needed on top, because v0.24.x only
   avoids the panic; it still stalls.

## Test plan

- [x] `go test ./submitter/...` — passes, including the 5 new cases.
- [x] `go test ./...` — full suite green (including e2e).
- [x] `go vet ./...` clean.
- [x] `gofmt -l ./...` clean.
- [ ] Manual testnet reproduction: submit a checkpoint, SIGKILL the
      submitter before tx2 confirms, restart, confirm logs show
      `Rehydrated lastSubmittedCheckpoint for epoch N from store` and
      that a subsequent RBF (if the fee window kicks in) succeeds
      instead of panicking / looping.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/submitter/relayer/relayer.go
+++ b/submitter/relayer/relayer.go
@@ -135,6 +135,15 @@ func (rl *Relayer) SendCheckpointToBTC(ckpt *ckpttypes.RawCheckpointWithMetaResp
 		}
 
 		if hasBeenProcessed {
+			// Restore in-memory state from the store so the subsequent RBF path
+			// in MaybeResubmitSecondCheckpointTx has valid Tx1/Tx2 info. Without
+			// this, a restart while a checkpoint is still in the mempool leaves
+			// lastSubmittedCheckpoint zero-valued and RBF attempts panic (pre
+			// v0.24.0-rc.3) or loop forever on a nil-Tx2 error.
+			if rl.lastSubmittedCheckpoint.Tx2 == nil {
+				rl.rehydrateLastSubmittedCheckpointFromStore(ckptEpoch)
+			}
+
 			return nil
 		}
 	}
@@ -181,6 +190,12 @@ func (rl *Relayer) MaybeResubmitSecondCheckpointTx(ckpt *ckpttypes.RawCheckpoint
 	if ckpt.Status != ckpttypes.Sealed {
 		rl.logger.Errorf("The checkpoint for epoch %v is not sealed", ckptEpoch)
 		rl.metrics.InvalidCheckpointCounter.Inc()
+
+		return nil
+	}
+
+	if rl.lastSubmittedCheckpoint.Tx1 == nil || rl.lastSubmittedCheckpoint.Tx2 == nil {
+		rl.logger.Debugf("lastSubmittedCheckpoint not populated for epoch %v, skipping RBF attempt", ckptEpoch)
 
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


submitter/relayer/relayer_test.go
```diff
@@ -1779,3 +1783,128 @@ func (m *MockCounter) Write(*prometheus.Metric) error {
...
+		Estimator:               &MockEstimator{},
+		store:                   subStore,
+		config:                  &config.SubmitterConfig{ResendIntervalSeconds: 10},
+		logger:                  logger.Sugar().With("module", "relayer"),
+		lastSubmittedCheckpoint: &types.CheckpointInfo{},
+	}
+}
+
+// nolint:paralleltest
+func TestRehydrateLastSubmittedCheckpointFromStore(t *testing.T) {
```
Lazar955, line 1807: nit; I prefer table tests

submitter/relayer/relayer.go
```diff
@@ -1027,6 +1039,93 @@ func maybeResendFromStore(
...
+// lastSubmittedCheckpoint from the persisted StoredCheckpoint after a
+// restart. Without this, a restart while the checkpoint is still in the
+// mempool leaves lastSubmittedCheckpoint zero-valued and the RBF path in
+// MaybeResubmitSecondCheckpointTx operates on nil Tx1/Tx2.
+//
+// Fee and TS aren't persisted in the store, so both are pulled from the
+// BTC node's mempool entry (Time for TS, so the resend interval reflects
+// actual mempool age rather than restart time). On lookup failure we fall
+// back to calcMinRelayFee + time.Now().
+func (rl *Relayer) rehydrateLastSubmittedCheckpointFromStore(ckptEpoch uint64) {
```
Lazar955: Should rehydrateLastSubmittedCheckpointFromStore return an error instead of silently swallowing failures?

Every failure path (store read, vsize computation, mempool lookup for tx1 or tx2) logs a Warn and returns, leaving lastSubmittedCheckpoint zero-valued. The caller at line 141 does not check whether rehydration succeeded and returns nil as if everything is fine. On the next poll cycle, MaybeResubmitSecondCheckpointTx hits the nil guard at line 194 and silently skips the RBF attempt for that epoch. This means a transient store or RPC error permanently disables fee-bumping until the next restart (or epoch rollover). Would returning an error let processCheckpoints retry on the next tick instead of silently degrading?

submitter/relayer/relayer.go
```diff
@@ -185,6 +191,12 @@ func (rl *Relayer) MaybeResubmitSecondCheckpointTx(ckpt *ckpttypes.RawCheckpoint
 		return nil
 	}
 
+	if rl.lastSubmittedCheckpoint.Tx1 == nil || rl.lastSubmittedCheckpoint.Tx2 == nil {
+		rl.logger.Debugf("lastSubmittedCheckpoint not populated for epoch %v, skipping RBF attempt", ckptEpoch)
```
Lazar955: let's do warn log, otherwise this is probably omitted in most envs

submitter/relayer/relayer.go
```diff
@@ -1027,6 +1039,93 @@ func maybeResendFromStore(
...
+	}
+
+	size, err := calculateTxVirtualSize(tx)
+	if err != nil {
+		return nil, time.Time{}, fmt.Errorf("failed to compute virtual size: %w", err)
+	}
+
+	txHash := tx.TxHash()
+
+	fee := rl.calcMinRelayFee(size)
```
Lazar955: Is calcMinRelayFee the right fallback when the mempool entry is unavailable?

If the tx was confirmed (mined between maybeResendFromStore and btcTxInfoFromStoredTx), the actual fee was likely higher than min-relay-fee. When calculateBumpedFee later computes bumpedFee = ckptInfo.Tx2.Fee * ResubmitFeeMultiplier, an artificially low Tx2.Fee could produce an RBF replacement that violates BIP125 rule 4 (new fee must be >= old fee + incremental relay fee). The replacement would be rejected by the node, producing a loop of failed RBF attempts until the original tx confirms. Would returning an error here (letting the next cycle re-evaluate) be safer than guessing a fee?

Lazar955 (approved):
Non blocking comments, thanks for the PR

PR #543, opened 2026-04-28
fix(btcslasher): durable bootstrap cursor via pending evidence store

Lazar955:
## Summary

- Fixes a BTC-side slashing bypass triggered by a vigilante restart between slashing dispatch and BTC tx broadcast. The bootstrap cursor advanced on async dispatch, so on restart Babylon's `ListEvidences(startHeight)` would filter out an FP whose first slashable evidence sat below the persisted cursor, leaving it permanently un-slashed.
- Adds a persistent `pendingevidences` bucket to the slasher kvdb. New evidences are written before slashing dispatch, then deleted by a per-FP wrapper goroutine once every per-delegation goroutine reaches a terminal state (BTC tx K-deep or all paths non-slashable). On graceful shutdown the entries are intentionally left in place so the next Bootstrap replays them.
- On Bootstrap, pending entries are replayed before the cursor-based query. Bitcoin tx submission is already idempotent via `isTxSubmittedToBitcoin`, so replay is safe even when a tx was previously broadcast.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Bug Fixes
 
-* [#537](https://github.com/babylonlabs-io/vigilante/pull/537) fix(btcslasher): durable bootstrap cursor via pending evidence store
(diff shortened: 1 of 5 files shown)


RafilxTenfen:
CI is failing @Lazar955

PR #548, opened 2026-05-05
test(e2e): un-skip GHSA stake-expansion test by overriding babylond image to v4.3.0

RafilxTenfen:
## Summary

`TestStakeExpansionParentUnbondAtOneDeepWhenChildUnbondedEarly` was added in #545 with a `t.Skip` because the assertion requires server-side behavior shipped in babylon `v4.3+`, while `main`'s `go.mod` still pins `babylon/v4 v4.2.x`. Bumping the Go dep is a larger change with broader implications; this PR introduces a smaller surgical alternative that lets the test run today.

## Approach

Decouple the babylond docker image tag from the `go.mod` version on a per-test basis:

- `container.Manager.SetBabylonVersion(version string)` — public setter so the image tag can be swapped after `NewManager` reads the default from `go.mod` via `testutil.GetBabylonVersion`.
- `TestManagerConfig.BabylonVersion` field + `WithBabylonVersion(version string)` test option, applied in `StartManager` before `RunBabylondResource` fires.
- The GHSA test drops its `t.Skip` and instead passes `WithBabylonVersion("v4.3.0")`, so it exercises the v4.3 server behavior while the Go SDK stays at v4.2.x.

The protobuf wire format must remain compatible between client and server — that's why the option is opt-in, not a global flip. Other e2e tests are unaffected; they keep using the `go.mod`-derived default.

## Origin

Backport of the equivalent change made on `release/v0.24.x` (PR #546). Same patch, just adjusted to reference v4.2.x in the comment instead of v4.0.0.

## Test plan

- [x] `go build -tags e2e ./e2etest/...`
- [ ] CI integration-tests green — the test now runs to completion against the v4.3.0 image and validates the GHSA-wcr8-g34v-7565 fix end-to-end

## Follow-up

When `babylon/v4` is bumped to `v4.3.x` in `go.mod`, drop the `WithBabylonVersion` override in this test (the default will catch up).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/e2etest/test_manager.go
+++ b/e2etest/test_manager.go
@@ -63,6 +63,9 @@ type TestManagerConfig struct {
 	NumMatureOutputsInWallet uint32
 	EpochInterval            uint
 	NumCovenants             uint
+	// BabylonVersion overrides the babylond docker image tag for this test.
+	// Empty means use the default derived from go.mod.
+	BabylonVersion string
 }
 
 func defaultTestManagerConfig() *TestManagerConfig {
@@ -88,6 +91,16 @@ func WithEpochInterval(interval uint) TestManagerOption {
 func WithNumCovenants(numCovenants uint) TestManagerOption {
 	return func(config *TestManagerConfig) {
 		config.NumCovenants = numCovenants
+	}
+}
+
+// WithBabylonVersion runs the babylond container at a specific image tag,
+// independent of the babylon Go module version pinned in go.mod. Use sparingly
+// — the protobuf wire format must remain compatible between the client (Go
+// module) and the server (image tag).
+func WithBabylonVersion(version string) TestManagerOption {
+	return func(config *TestManagerConfig) {
+		config.BabylonVersion = version
 	}
 }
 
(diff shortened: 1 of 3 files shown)


e2etest/test_manager.go
```diff
@@ -91,6 +94,16 @@ func WithNumCovenants(numCovenants uint) TestManagerOption {
 	}
 }
 
+// WithBabylonVersion runs the babylond container at a specific image tag,
+// independent of the babylon Go module version pinned in go.mod. Use sparingly
+// — the protobuf wire format must remain compatible between the client (Go
+// module) and the server (image tag).
+func WithBabylonVersion(version string) TestManagerOption {
```
Lazar955, line 101: thanks for respecting the pre llm vision :D

e2etest/unbondingwatcher_e2e_test.go
```diff
@@ -929,18 +929,21 @@ func TestStakeExpansionFlow(t *testing.T) {
...
+	// The end-to-end assertion requires babylond to accept the parent's
+	// MsgBTCUndelegate at sub-k depth when the child stake-expansion is
+	// already unbonded — that server-side change ships in babylon v4.3+.
+	// go.mod still pins v4.2.x for the Go SDK; we only override the docker
+	// image tag so this test exercises the v4.3 server behavior. Drop the
+	// override once the babylon Go dep is bumped.
+	tm := StartManager(t,
+		WithNumMatureOutputs(numMatureOutputs),
+		WithEpochInterval(defaultEpochInterval),
+		WithBabylonVersion("v4.3.0"),
```
Lazar955: nit, maybe const tho not improtant

Lazar955 (approved):
thanks! Adding BP label too

PR #551, opened 2026-05-22
fix(reporter): keep checkpoint segments until Babylon accepts the proof

Lazar955:
## Summary

 Previously `CheckpointCache.Match()` deleted both checkpoint segments as soon as a `part0`/`part1` pair connected and decoded locally, before Babylon validated the proof. An attacker who observes a legitimate `part0` on Bitcoin can forge a `part1` with the correct 10-byte checksum but a corrupted BLS signature, get it mined before the real `part1`, and cause the reporter to consume the real `part0` on a proof that Babylon rejects. The orphaned real `part1` can then never match.

## Fix

- `Match()` no longer deletes segments at match time. It queues candidate checkpoints and skips pairs already recorded in a new `attemptedPairs` set.
- Submission outcome now drives segment lifecycle in `matchAndSubmitCheckpoints`:
  - success / expected (`ErrDuplicatedSubmission`, `ErrEpochAlreadyFinalized`) → `RemoveSegments`
  - transient (`ErrTimeoutAfterWaitingForTxBroadcast`) → leave segments untouched for retry
  - other rejection (e.g. invalid BLS) → `MarkAttempted`, keeping `part0` cached so a later legitimate `part1` can still match, while blacklisting the rejected pair so the same proof isn't resubmitted
- Cleanup routine prunes `attemptedPairs` on the existing segment TTL.

## Tests

- `types/ckpt_cache_poison_test.go`: cache-level regression (steady-state ordering, dedup across repeated `Match()`, bootstrap recovery with both pairs queued, attempted-pair cleanup on success)
- `reporter/checkpoint_cache_poison_test.go`: reporter-level with mocked Babylon (poison recovery, dedup across invocations, transient-error retry)
- `e2etest/reporter_checkpoint_cache_poison_e2e_test.go`: full bitcoind/electrs/babylond stack; mines poison block, asserts Babylon rejects, mines real `part1`, asserts checkpoint advances past Sealed
- Updated `FuzzCheckpointCache` for the new no-delete-on-match contract

All touched-package tests pass; `go vet` and `golangci-lint` clean including `-tags=e2e`.

### DIFF
--- a/reporter/utils.go
+++ b/reporter/utils.go
@@ -320,6 +320,18 @@ func (r *Reporter) matchAndSubmitCheckpoints(signer string) int {
 					tx1Block.Txs[ckpt.Segments[0].TxIdx].Hash().String(),
 					tx2Block.Txs[ckpt.Segments[1].TxIdx].Hash().String(),
 				).Inc()
+
+				// Transient broadcast timeout: leave segments in the cache
+				// without marking the pair as attempted, so the next Match()
+				// will retry the same pair.
+			} else {
+				// Non-transient rejection (e.g. invalid BLS multi-sig from a
+				// poisoned part1, or any other validation error). Keep the
+				// segments cached so the legitimate part1, when it arrives,
+				// can still pair with this part0 (VIG-02). But mark this
+				// specific (part0, part1) pair as attempted so subsequent
+				// Match() calls do not resubmit the same rejected proof.
+				r.checkpointCache.MarkAttempted(ckpt)
 			}
 
 			continue
@@ -333,6 +345,11 @@ func (r *Reporter) matchAndSubmitCheckpoints(signer string) int {
 			r.logger.Infof("Successfully submitted checkpoint (MsgInsertBTCSpvProof) with response %d, for epoch %d, height %d",
 				res.Code, ckpt.Epoch, tx1Block.Height)
 		}
+
+		// Babylon accepted the proof (or returned an expected duplicate /
+		// already-finalized response). Drop the source segments from the
+		// cache so the pair is not re-matched on the next Match() cycle.
+		r.checkpointCache.RemoveSegments(ckpt)
 
 		// either way if we or some other reporter submitted the checkpoint, we want to update the metrics
 		r.metrics.SuccessfulCheckpointsCounter.Inc()
--- a/types/ckpt_cache.go
+++ b/types/ckpt_cache.go
@@ -22,6 +22,13 @@ type CheckpointCache struct {
 	// first key: index of the segment in the checkpoint (0 or 1)
 	// second key: hash of the OP_RETURN data in this ckpt segment
 	Segments map[uint8]map[string]*CkptSegment
+
+	// attemptedPairs tracks (hash(part0) | hash(part1)) keys for matched pairs
+	// that have already been submitted to Babylon and rejected with a
+	// non-transient error. Match() skips these so the same poisoned proof
+	// is not resubmitted on every cycle. Entries are pruned by the cleanup
+	// routine using the same TTL applied to segments.
+	attemptedPairs map[string]time.Time
 }
 
 func NewCheckpointCache(tag btctxformatter.BabylonTag, version btctxformatter.FormatVersion) *CheckpointCache {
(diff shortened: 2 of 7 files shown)


types/ckpt_cache.go
```diff
@@ -63,30 +71,50 @@ func (c *CheckpointCache) sortCheckpoints() {
...
+// rejected by Babylon (see VIG-02 fix).
 func (c *CheckpointCache) Match() {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	for hash1, ckptSeg1 := range c.Segments[uint8(0)] {
-		for hash2, ckptSeg2 := range c.Segments[uint8(1)] {
+	for _, ckptSeg1 := range c.Segments[uint8(0)] {
+		for _, ckptSeg2 := range c.Segments[uint8(1)] {
+			if _, ok := c.attemptedPairs[pairKey(ckptSeg1, ckptSeg2)]; ok {
```
RafilxTenfen, line 99:
```suggestion
  for hash1, ckptSeg1 := range c.Segments[uint8(0)] {
      for hash2, ckptSeg2 := range c.Segments[uint8(1)] {
          if _, ok := c.attemptedPairs[hash1+"|"+hash2]; ok {

```

reporter/utils.go
```diff
@@ -320,6 +320,18 @@ func (r *Reporter) matchAndSubmitCheckpoints(signer string) int {
...
+				// without marking the pair as attempted, so the next Match()
+				// will retry the same pair.
+			} else {
+				// Non-transient rejection (e.g. invalid BLS multi-sig from a
+				// poisoned part1, or any other validation error). Keep the
+				// segments cached so the legitimate part1, when it arrives,
+				// can still pair with this part0 (VIG-02). But mark this
+				// specific (part0, part1) pair as attempted so subsequent
+				// Match() calls do not resubmit the same rejected proof.
+				r.checkpointCache.MarkAttempted(ckpt)
```
RafilxTenfen, line 334:
```suggestion
  } else {
      // Only blacklist when the rejection identifies the proof as
      // structurally bad (the VIG-02 poison case). Network / RPC failures
      // and module errors that aren't proof-validation failures fall
      // through and get retried on the next Match() cycle.
      if isInvalidProofError(err) {
          r.checkpointCache.MarkAttempted(ckpt)
      }
  }

```

RafilxTenfen: claude suggestion: 
```
the else branch treats anything that isn't ErrTimeoutAfterWaitingForTxBroadcast as a permanent rejection. That bucket includes transient Babylon-RPC failures (network down, gRPC
  Unavailable, deadline-exceeded that isn't the broadcast wait, transaction failed with code: %d for non-BLS module errors). A legitimate pair submitted during a Babylon outage gets
  blacklisted for the full segment TTL (24h) and is unrecoverable without a restart.

```

types/ckpt_cache.go
```diff
@@ -108,6 +136,39 @@ func (c *CheckpointCache) PopEarliestCheckpoint() *Ckpt {
...
+		return
+	}
+	h0 := sha256.Sum256(ckpt.Segments[0].Data)
+	h1 := sha256.Sum256(ckpt.Segments[1].Data)
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	delete(c.Segments[uint8(0)], string(h0[:]))
+	delete(c.Segments[uint8(1)], string(h1[:]))
+	delete(c.attemptedPairs, string(h0[:])+"|"+string(h1[:]))
+}
```
RafilxTenfen, line 170: if that previous suggestion is applied
```suggestion
  func (c *CheckpointCache) RemoveSegments(ckpt *Ckpt) {
      if ckpt == nil || len(ckpt.Segments) != int(btctxformatter.NumberOfParts) {
          return
      }
      h0 := sha256.Sum256(ckpt.Segments[0].Data)
      h1 := sha256.Sum256(ckpt.Segments[1].Data)
      k0, k1 := string(h0[:]), string(h1[:])
      c.mu.Lock()
      defer c.mu.Unlock()
      delete(c.Segments[uint8(0)], k0)
      delete(c.Segments[uint8(1)], k1)
      delete(c.attemptedPairs, k0 + "|" + k1)
  }

```

PR #556, opened 2026-09-17
fix(submitter): charge only RBF fee delta on resend and commit replacement after broadcast

### LINKED ISSUE #552: Submitter RBF resend overpays miner fee by subtracting full bumpedFee from already fee-reduced change output
## Summary

In `maybeResendSecondTxOfCheckpointToBTC` (submitter/relayer/relayer.go), the existing-input branch reduces the change output by the full `bumpedFee` rather than by the delta `bumpedFee - oldFee`. Because the change output has already had the original fee deducted at funding time, the broadcast transaction pays `oldFee + bumpedFee` in actual miner fees while the relayer records only `bumpedFee` in `tx2.Fee`. Each RBF resend therefore overpays by approximately one previous-cycle fee.

A secondary defect: the transaction is mutated in place before signing and broadcasting. If `SendRawTransaction` fails, the in-memory `tx2.Tx` keeps the reduced change output while `tx2.Fee` stays at the old value, so a subsequent retry compounds the overpayment.

## Affected code

submitter/relayer/relayer.go, function `maybeResendSecondTxOfCheckpointToBTC`, lines 491 to 524:

```go
balance := btcutil.Amount(tx2.Tx.TxOut[changePosition].Value)
originalTxID := tx2.Tx.TxID()

if balance-bumpedFee < dustThreshold {
    // dust path: rebuilds via FundRawTransaction, correct fee accounting
    ...
} else {
    // bug: subtracts full bumpedFee from already fee-reduced change
    tx2.Tx.TxOut[changePosition].Value = int64(balance - bumpedFee)
}
```

The mutation happens before `verifyRBFRequirements`, `signTx`, and `sendTxToBTC`. `tx2.Fee = bumpedFee` is only assigned on the success path at line 543.

## Fee math example

Input value 100,000 sats, original change 90,000 sats, original fee 10,000 sats, target replacement fee 20,000 sats.

Correct replacement:
- new change = 100,000 - 20,000 = 80,000 sats
- actual fee = 20,000 sats

Current behavior:
- new change = 90,000 - 20,000 = 70,000 sats
- actual fee = 100,000 - 70,000 = 30,000 sats
- recorded fee = 20,000 sats
- overpay = 10,000 sats (one previous-cycle fee)

## Production impact (mainnet measurement, 30-day window)

Measured via Prometheus and Loki on the mainnet submitter pod:

| Metric | Value |
|---|---|
| `vigilante_submitter_resent_checkpoints` | ~17 successful resends per day |
| `vigilante_submitter_failed_resent_checkpoints` (7d) | 0 |
| Per-resend overpay (median) | ~420 sats |
| Cumulative overpay (7d rolling, median) | ~40,000 sats |
| Annualized overpay | ~2.08M sats, roughly 0.021 BTC |

At a $78,000 BTC reference price, the production cost is approximately $1,600 per year of overpaid miner fees from the operator hot wallet. The compounding retry case has not fired in the last 7 days on mainnet.

## Reproduction

Two table-driven tests in `submitter/relayer/rbf_fee_overpay_poc_test.go` 
demonstrate both the single-success overpay and the failed-then-retry compounding case. Both pass against current `release/v0.24.x`:

```
cd vigilante
go test -v -timeout 60s ./submitter/relayer -run '^TestPoC_' -count=1
```

Expected output:
```
--- PASS: TestPoC_RBFFeeBumpSubtractsTotalFeeAndOverpays
--- PASS: TestPoC_FailedRBFBroadcastCompoundsOverpayment
```

## Fix plan

1. Replace the full-fee subtraction in the existing-input branch with a delta-based update:

   ```go
   if bumpedFee <= tx2.Fee {
       return nil, fmt.Errorf("replacement fee must exceed old fee: bumped=%v old=%v", bumpedFee, tx2.Fee)
   }
   feeDelta := bumpedFee - tx2.Fee
   newChangeValue := int64(balance - feeDelta)
   ```

2. Mutate a copy of `tx2.Tx`, not the in-memory original. Assign the copy back into `tx2.Tx` only after `signTx` and `sendTxToBTC` both succeed. This makes failed broadcast attempts idempotent and eliminates the compounding path even though it is not firing in production today.

3. Add a regression test that asserts actual on-chain fee (input value minus all output values) equals `bumpedFee` after the resend, not `oldFee + bumpedFee`.

4. While touching this code, audit `verifyRBFRequirements` and the `adjustFeeFor*` helpers for the `btcutil.Amount(float)` versus `btcutil.NewAmount(float)` issue on mempool JSON fee fields (`mempoolEntry.DescendantFees`, `mempoolEntry.Fee`). Direct `btcutil.Amount` casts on BTC-denominated floats truncate to near zero.

## Severity

Informational. Operator hot wallet only, no user funds, no attacker-induced trigger, bounded annualized cost on the order of $1.6k per year at current BTC price. 

## Acceptance criteria

- [ ] Delta-based change-output update lands in `maybeResendSecondTxOfCheckpointToBTC`
- [ ] Transaction mutation happens on a copy and is committed only after successful broadcast
- [ ] Regression test asserts actual on-chain fee equals `bumpedFee` for the existing-input branch
- [ ] Reporter's PoC tests pass with assertions inverted (i.e., overpay no longer occurs)
- [ ] `btcutil.NewAmount` audit completed for mempool JSON fee fields

Tyagiquamar:
Fixes #552. In maybeResendSecondTxOfCheckpointToBTC, the existing-input branch subtracted the full replacement fee from a change output that already had the original fee deducted at funding time, so each RBF resend paid oldFee plus bumpedFee while recording only bumpedFee. The mutation also happened in place before signing and broadcasting, so a failed broadcast compounded the overpay on retry. This changes the branch to subtract only bumpedFee minus tx2.Fee from change (erroring when the replacement fee does not exceed the recorded fee), builds the replacement on a copy, and commits it to tx2 only after signTx and sendTxToBTC succeed. Existing resend test expectations updated to the delta accounting; TestRelayer_MaybeResendSecondTxFeeDelta added for the single-resend fee and failed-broadcast idempotency. Verified in golang:1.25-bookworm: gofmt clean, go vet plus go build pass, targeted tests pass, full ./submitter/... suite passes; the new test fails on unpatched code.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

